Issue #67, opened 2020-09-04
As developer I want to have history of objects 


annashamray:
API should have two endpoints:
* `objects`: shows objects with the current record (today is between start_date and end_date)
* `history`: show objects with all records 

Document with examples of material and formal history: https://www.gemmaonline.nl/images/gemmaonline/f/fa/Stuf0301.pdf

PR #68, opened 2020-09-04
Feature/history of objects

annashamray:
fixes https://github.com/maykinmedia/objects-api/issues/67


src/objects/api/serializers.py
```diff
@@ -1,47 +1,78 @@
...
+        }
 
 
 class ObjectSerializer(serializers.HyperlinkedModelSerializer):
+    record = ObjectRecordSerializer(source="current_record")
+
     class Meta:
         model = Object
-        fields = ("url", "type", "version", "data")
+        fields = ("url", "type", "typeVersion", "record")
```
joeribekker, line 53: This comment will touch several places, but in essence:

* `type` should be readonly, it doesn't make sense to change the type.
* `typeVersion` should be part of the record. People need to know the version for each record.
* can we also add the `uuid` field?

joeribekker (changes requested):

src/objects/api/serializers.py
annashamray: `type` is immutable now, i.e. it can be created, but it can't be updated with a new value
`typeVersion` is moved to the record 
`uuid` field is added to Record resource and now in the API records are referenced with their uuid

joeribekker (approved):

PR #74, opened 2020-10-02
Feature/remove required geo headers

annashamray:
This is a work-around to be able to create `ZAAKOBJECT` with Object API urls in Zaken API

### DIFF
--- a/src/objects/api/views.py
+++ b/src/objects/api/views.py
@@ -17,7 +17,7 @@ from .serializers import (
 )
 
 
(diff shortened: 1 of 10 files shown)


joeribekker (approved):
In contrast to the Zaken API, I think geo-headers should be optional anyway for GET-requests. The response should ofcourse always have the header to show the format of the returned values.

When sending data (POST/PUT/PATCH), the header should be mandatory to prevent incorrect values.

I made #75 to address this.

PR #86, opened 2020-10-13
Feature/object type permissions

### LINKED ISSUE #78: As admin-user I want to set up permissions based on objects
- [x] Object permission should include `object_type` and read/write option
- [x] Each user can have only one object permission for each `object_type`

annashamray:
Fixes #78


joeribekker (approved):
Looks good, tests seem complete.

PR #87, opened 2020-10-20
Feature/filter on data

### LINKED ISSUE #34: As developer, I want to filter on **any** data-attribute

annashamray:
Fixes #34 

Implementation is inspired by Camunda REST API (https://docs.camunda.org/manual/7.11/reference/rest/task/get-query/ `taskVariables` parameter)

### DIFF
--- a/src/objects/api/filters.py
+++ b/src/objects/api/filters.py
@@ -6,14 +6,9 @@ from vng_api_common.utils import get_help_text
 
 from objects.core.models import Object
 
-
-def is_number(value: str) -> bool:
-    try:
-        float(value)
-    except ValueError:
-        return False
-
-    return True
+from .constants import Operators
+from .utils import display_choice_values_for_help_text, is_number
+from .validators import validate_data_attrs
 
 
 class ObjectFilterSet(FilterSet):
@@ -22,6 +17,7 @@ class ObjectFilterSet(FilterSet):
     )
     data_attrs = filters.CharFilter(
         method="filter_data_attrs",
+        validators=[validate_data_attrs],
         help_text=_(
             """Only include objects that have attributes with certain values.
 Data filtering expressions are comma-separated and are structured as follows:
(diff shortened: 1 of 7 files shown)


src/objects/api/filters.py
```diff
@@ -1,15 +1,67 @@
...
+Note: Values can be string or numeric. Dates are not supported.
+
+Valid operator values are:
+%(operator_choices)s
+`value` may not contain double underscore or comma characters.
+`key` may not contain comma characters and includes double underscore only if it indicates nested attributes.
+
+Example: in order to display only objects with `height` equal to 100, query `data_attrs=height__exact__100`
+should be used. If `height` is nested inside `dimensions` attribute, query should look like
+`data_attrs=dimensions__height__exact__100`
```
joeribekker, line 35: Is it possible to provide the query param `data_attrs` multiple times? And if yes, what is the effect of this?

src/objects/tests/test_filters.py
```diff
@@ -27,3 +27,139 @@ def test_filter_object_type(self):
...
+            response.json(), ["Operator `lt` supports only numeric values"]
+        )
+
+    def test_filter_invalid_operator(self):
+        response = self.client.get(self.url, {"data_attrs": "diameter__not__value"})
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json(), ["Comparison operator `not` is unknown"])
+
+    def test_filter_nester_attr(self):
```
joeribekker, line 124:
```suggestion
    def test_filter_nested_attr(self):
```

src/objects/api/filters.py
```diff
@@ -1,15 +1,67 @@
...
+from .constants import Operators
+from .utils import display_choice_values_for_help_text, is_number
+from .validators import validate_data_attrs
+
 
 class ObjectFilterSet(FilterSet):
     type = filters.CharFilter(
         field_name="object_type", help_text=get_help_text("core.Object", "object_type")
     )
+    data_attrs = filters.CharFilter(
```
joeribekker, line 18: I think this is a pretty smart way to do it but i differs alot with regular API query params.

I would prefer typical `key=value` pairs. Can we perhaps prefix query params for the `records.data` with `records__data__` so it looks like normal query params? Something like:

`?record__data__height__exact=100` and `?record__data__dimensions__height__exact=100`

src/objects/api/filters.py
```diff
@@ -1,15 +1,67 @@
...
+        method="filter_data_attrs",
+        validators=[validate_data_attrs],
+        help_text=_(
+            """Only include objects that have attributes with certain values.
+Data filtering expressions are comma-separated and are structured as follows:
```
joeribekker, line 23: I think the `data_attrs` (or `record__data__*` params should only be allowed if a `type` filter is provided, otherwise the results become messy.

joeribekker (changes requested):
I think this is pretty awesome but I wonder if it should be done slightly different, see comments.

src/objects/api/filters.py
joeribekker: Moved to #90

src/objects/api/filters.py
joeribekker: Moved to #90

src/objects/api/filters.py
annashamray: In this case only the last query param is applied. I suppose it's a default behaviour for django filters, for example, if you provide `type` query params multiple times, only the last param would be applied

PR #111, opened 2020-12-02
Updated layout

joeribekker:


annashamray (commented):
I have the same comments as for https://github.com/maykinmedia/objecttypes-api/pull/23

annashamray (approved):

PR #112, opened 2020-12-03
Updated demo data

joeribekker:
Followup on #111.


annashamray (changes requested):
zgw_consumers app is not in the INSTALLED_APPS, and its fixture can't be loaded

annashamray (approved):

PR #153, opened 2021-02-17
Feature/pagination

### LINKED ISSUE #148: As developer, I want the API to be paginated
Since the [API strategy is in development regarding pagination](https://docs.geostandaarden.nl/api/API-Strategie-ext/#paging), the example given there is also incorrect and using HAL, we revert to the method used by VNG for the Zaken API.

annashamray:
fixes #148


joeribekker (commented):
Approved, but this is a breaking change so, let's merge it in after sprint 7 and release 1.1.0 just before that.

annashamray:
the change is implemented at https://github.com/maykinmedia/objects-api/pull/205

Issue #155, opened 2021-02-17
As admin, I want to see more user information about an API token.


joeribekker:
This goes for both Objects API and Objecttypes API.

- [x] The token field should be read only
- [x] Add application field (optional)
- [x] Add administration field (optional)
- [x] Show organization, administrtion, application and contact person also in the list view

PR #164, opened 2021-03-12
Feature/token add fields

annashamray:
Issue #155


src/objects/token/management/commands/generate_token.py
```diff
@@ -21,14 +21,30 @@ def add_arguments(self, parser):
             default="",
             help=_("Name of the organization"),
         )
+        parser.add_argument(
```
joeribekker, line 24: Nice, I forgot about this command when this user story was made. Good catch.

joeribekker (approved):

src/objects/token/models.py
```diff
@@ -33,6 +33,18 @@ class TokenAuth(models.Model):
     created = models.DateTimeField(
         _("created"), auto_now_add=True, help_text=_("Date when the token was created")
     )
+    application = models.CharField(
+        _("application"),
+        max_length=200,
+        blank=True,
+        help_text=_("Application which has an access to the API"),
```
joeribekker, line 40: > which has an access

The "an" should be left in all 3 help_texts.

annashamray: done

PR #166, opened 2021-03-15
Feature/record optimization

### LINKED ISSUE #136: As developer, I want to optimize the record retrieval for faster performance
Currently, we use the `to_representation` method to swap in the correct Object record, based on query params. This is probably not the best way to approach this (although functionally it works!).

See: https://github.com/maykinmedia/objects-api/pull/134/files#diff-85bba2c9e2f4672c7a1c89e8c7ee8b95d4c3e54aeab953b3f7ede8e4288d19f8R92

In essence, the `ObjectSerializer` should pass the proper `Record` instance to the `ObjectRecordSerializer`.

annashamray:
fixes #136

### DIFF
--- a/src/objects/conf/base.py
+++ b/src/objects/conf/base.py
@@ -227,7 +227,7 @@ LOGGING = {
         "django": {
             "level": "DEBUG",
             "class": "logging.handlers.RotatingFileHandler",
-            "filename": os.path.join(LOGGING_DIR, "django.log"),
+            "filename": os.path.join(LOGGING_DIR, "django.log2"),
             "formatter": "verbose",
             "maxBytes": 1024 * 1024 * 10,  # 10 MB
             "backupCount": 10,
--- a/src/objects/conf/base.py
+++ b/src/objects/conf/base.py
@@ -227,7 +227,7 @@ LOGGING = {
         "django": {
             "level": "DEBUG",
             "class": "logging.handlers.RotatingFileHandler",
-            "filename": os.path.join(LOGGING_DIR, "django.log2"),
+            "filename": os.path.join(LOGGING_DIR, "django.log"),
             "formatter": "verbose",
             "maxBytes": 1024 * 1024 * 10,  # 10 MB
             "backupCount": 10,
--- a/src/objects/core/models.py
+++ b/src/objects/core/models.py
@@ -11,7 +11,7 @@ from requests.exceptions import ConnectionError
 from zds_client.client import ClientError
 from zgw_consumers.models import Service
 
-from .query import ObjectQuerySet, ObjectTypeQuerySet
+from .query import ObjectQuerySet, ObjectRecordQuerySet, ObjectTypeQuerySet
 from .utils import check_objecttype
 
 
@@ -62,25 +62,9 @@ class Object(models.Model):
 
     objects = ObjectQuerySet.as_manager()
 
-    def get_actual_record(self, date):
-        return (
-            self.records.filter(start_at__lte=date)
-            .filter(models.Q(end_at__gte=date) | models.Q(end_at__isnull=True))
-            .order_by("-pk")
-            .first()
-            # TODO: pk should prolly be index once added.
-        )
-
-    def get_registration_record(self, date):
-        return (
-            self.records.filter(registration_at__lte=date)
-            .order_by("-registration_at")
-            .first()
-        )
-
     @property
     def current_record(self):
-        return self.get_actual_record(datetime.date.today())
+        return self.records.filter_for_date(datetime.date.today()).first()
 
     @property
     def last_record(self):
@@ -128,6 +112,8 @@ class ObjectRecord(models.Model):
             "Point, linestring or polygon object which represents the coordinates of the object"
         ),
     )
+
+    objects = ObjectRecordQuerySet.as_manager()
 
     class Meta:
         unique_together = ("object", "index")
--- a/src/objects/core/models.py
+++ b/src/objects/core/models.py
@@ -68,7 +68,7 @@ class Object(models.Model):
 
     @property
     def last_record(self):
-        return self.records.order_by("-start_at", "-id").first()
+        return self.records.order_by("-start_at", "-index").first()
 
 
 class ObjectRecord(models.Model):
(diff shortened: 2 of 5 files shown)


src/objects/conf/base.py
```diff
@@ -227,7 +227,7 @@
         "django": {
             "level": "DEBUG",
             "class": "logging.handlers.RotatingFileHandler",
-            "filename": os.path.join(LOGGING_DIR, "django.log"),
+            "filename": os.path.join(LOGGING_DIR, "django.log2"),
```
joeribekker, line 230: Hmm?

joeribekker (changes requested):
Awesome!

This cut down retrieval time with 2500 objects from 50s to 1.5s :)

Only the log change is not needed.

annashamray:
> This cut down retrieval time with 2500 objects from 50s to 1.5s :)

@joeribekker cool, thanks for testing it on a large number of objects. And I think pagination will also reduce the request time ;)

src/objects/core/models.py
```diff
@@ -62,25 +62,9 @@ class Object(models.Model):
...
-            .first()
-        )
-
     @property
     def current_record(self):
-        return self.get_actual_record(datetime.date.today())
+        return self.records.filter_for_date(datetime.date.today()).first()
 
     @property
     def last_record(self):
```
joeribekker, line 70: Shouldn't the last record be indicated simply with:

```
self.records.order_by("-index").first()
```

?

src/objects/core/query.py
```diff
@@ -32,3 +30,17 @@ def filter_for_date(self, date):
 
     def filter_for_registration_date(self, date):
         return self.filter(records__registration_at__lte=date).distinct()
+
+
+class ObjectRecordQuerySet(models.QuerySet):
+    def filter_for_date(self, date):
```
joeribekker, line 36:
```suggestion
    def filter_for_date(self, date):
        """
        Return records as seen on `date` from a formal historical perspective.

        The records that have their `start_at` date and `end_at` date between 
        the given `date`. If there is no `end_at` date, it means the record is
        still actual.
        
        If there are multiple records returned, the last added record (ie. with 
        the highest `index`) is the most actual record from a formal historical
        perspective.
        """
```

src/objects/core/query.py
```diff
@@ -32,3 +30,17 @@ def filter_for_date(self, date):
...
+        return (
+            self.filter(start_at__lte=date)
+            .filter(models.Q(end_at__gte=date) | models.Q(end_at__isnull=True))
+            .order_by("-index")
```
joeribekker, line 40: The `order_by` is missing in `ObjectQuerySet` counterpart above?

src/objects/core/query.py
```diff
@@ -32,3 +30,17 @@ def filter_for_date(self, date):
...
+        )
+
+    def filter_for_registration_date(self, date):
+        return self.filter(registration_at__lte=date).order_by(
+            "-registration_at", "-index"
```
joeribekker, line 45: The `order_by` is missing in `ObjectQuerySet` counterpart above?

src/objects/core/query.py
```diff
@@ -32,3 +30,17 @@ def filter_for_date(self, date):
...
+
+class ObjectRecordQuerySet(models.QuerySet):
+    def filter_for_date(self, date):
+        return (
+            self.filter(start_at__lte=date)
+            .filter(models.Q(end_at__gte=date) | models.Q(end_at__isnull=True))
+            .order_by("-index")
+        )
+
+    def filter_for_registration_date(self, date):
```
joeribekker, line 43:
```suggestion
    def filter_for_registration_date(self, date):
        """
        Return records as seen on `date` and later, from a material historical 
        perspective.

        Typically, the first record in the result set represents the record as 
        it is most actual from a material historical perspective.
        """
```

joeribekker (changes requested):

src/objects/core/query.py
annashamray: `ObjectRecordQuerySet.filter_for_registration_date` has sorting since we extract the first record later. `ObjectQuerySet.filter_for_registration_date` filters objects themselves not their records, and it is used in `ObjectViewSet`, where queryset is already sorted.

src/objects/core/query.py
annashamray: `ObjectRecordQuerySet.filter_for_date` has sorting since we extract the first record later. `ObjectQuerySet.filter_for_date` filters objects themselves not their records, and it is used in `ObjectViewSet`, where queryset is already sorted.

src/objects/core/models.py
joeribekker: > I thought that you can create a record that would start in the future, in this case, it wouldn't be retrieved here.

True, but is that the goal of this function? You only use `last_record` in the `ObjectRecord.save` method (as far as I can tell):

```python
    def save(self, *args, **kwargs):
        if not self.id and self.object.last_record:
            self.index = self.object.last_record.index + 1

            #  add end_at to previous record
            previous_record = self.object.last_record
            previous_record.end_at = self.start_at
            previous_record.save()

        super().save(*args, **kwargs)
```
 As for:

```python
        if not self.id and self.object.last_record:
            self.index = self.object.last_record.index + 1
```

The idea is that every record has an index value in the order they were added to the system, right? If you take `start_at` into account this can mess up things.

```python
            previous_record = self.object.last_record
            previous_record.end_at = self.start_at
            previous_record.save()
```

Here, it's different and I wonder if this is correct. If I correct a previous record, should the previous record then be adjusted?

annashamray: Yeah, you are right, I'll merge `current_record` and `last_record` together into one property that displays the record with the maximum index.

About corrections: we don't take them into account when displaying objects and their records, and we don't validate `start_date` and `end_date` of correcting records. Perhaps we need to think over how corrections should work and perhaps ask stakeholders?
Questions from the top of my head:
* could correcting record have `start_date` different from the correcred record
* if yes (and in StuF it's possible) should we change `end_date` of the record before the corrected one
* If they have different `start_date` it is possible that we will show the corrected record in the list endpoint. Is it ok or we need to exclude corrected records from being displayed?

joeribekker: As discussed, we'll leave this as is and do the rest in #168

joeribekker (approved):

PR #167, opened 2021-03-16
Changed pk to index order and added docstr

joeribekker:
I found some obsolete ordering by PK. I think this should be index and never date related.


joeribekker:
Added work as comments in #166

PR #186, opened 2021-03-31
Feature/helm chart

### LINKED ISSUE #180: As admin i want to have helm charts for deployment in k8s cluster
Helm charts should be created for:
* [x] Objects API - #186 
* [x] Objecttypes API - https://github.com/maykinmedia/objecttypes-api/pull/35

annashamray:
fixes #180


deployment/kubernetes/objects/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v2
+name: objects
+description: A Helm chart for Objects API
+
+type: application
+version: 0.1.0
```
joeribekker, line 6: Should this match our Docker container version? If so, should it be added to bumpversion config?

deployment/kubernetes/README.md
```diff
@@ -0,0 +1,31 @@
+# Objects chart
+Here you can find a reference implementation of Objects deployment in the
+Kubernetes cluster via [Helm](https://helm.sh/).
+This Helm chart installs Objects API and is dependent on [PostgreSQL](https://github.com/bitnami/charts/tree/master/bitnami/postgresql)
+subchart.
+
+:warning: The default settings are unsafe and should be used only for development purposes.
```
joeribekker, line 7:
```suggestion
:warning: The default settings are unsafe and should only be used for development purposes.
```

deployment/kubernetes/README.md
```diff
@@ -0,0 +1,31 @@
+# Objects chart
+Here you can find a reference implementation of Objects deployment in the
+Kubernetes cluster via [Helm](https://helm.sh/).
+This Helm chart installs Objects API and is dependent on [PostgreSQL](https://github.com/bitnami/charts/tree/master/bitnami/postgresql)
+subchart.
```
joeribekker, line 5:
```suggestion

Here you can find a reference implementation of the Objects API deployment for
a Kubernetes cluster using [Helm](https://helm.sh/).

This Helm chart installs the Objects API and is dependent on a [PostgreSQL](https://github.com/bitnami/charts/tree/master/bitnami/postgresql)
database, installed using a subchart.
```

deployment/kubernetes/README.md
```diff
@@ -0,0 +1,31 @@
...
+helm dependency build .
+helm install objects .
+```
+
+Use Kubernetes CLI to monitor the status of deployment:
+```bash
+kubectl get pods
+```
+
+If ingress is not configured you can use `port-forward` to check the status of application:
```
joeribekker, line 25:
```suggestion
If the Ingress is not configured you can use `port-forward` to check the status of the application:
```

joeribekker (changes requested):
As basis ok, small remarks.

deployment/kubernetes/objects/Chart.yaml
```diff
@@ -0,0 +1,14 @@
...
+appVersion: "1.16.0"
+
+dependencies:
+  - name: postgresql
+    version: ~10.2.0
```
joeribekker, line 11: Shouldn't we use a newer version?

joeribekker (changes requested):

deployment/kubernetes/objects/Chart.yaml
annashamray: Afaik this is a helm chart version and it should be incremented when the helm chart itself is changed
Docker container is deployed with "latest" tag now, which is specified in `value.yaml` file

deployment/kubernetes/objects/Chart.yaml
annashamray: Our `INSTALL.rst` file requires Postgres 10 or above and our ci also used Postgres 10 to run the test suite.
Tbh I don't want to touch the helm chart again until we will stop supporting Postgres 10 for this project

joeribekker: Ok.

joeribekker (approved):

Issue #195, opened 2021-05-17
As developer, I want to have access to v1 and v2 of the API


joeribekker:
Where version 2 has pagination, and version 1 has not (for example).

PR #201, opened 2021-06-07
Feature/drf spectacular

annashamray:
Migrate from `drf-yasg` to `drf-spectacular`

Custom error responses are removed from the schema, because Objects API doesn't have the same `exception_handler` as Zaken API and doesn't actually return responses with this format. 

Issue - #195  (3rd step)


src/objects/utils/autoschema.py
```diff
@@ -1,31 +1,111 @@
...
+                    description=_(
+                        "The 'Coordinate Reference System' (CRS) of the request data. "
+                        "According to the GeoJSON spec, WGS84 is the default (EPSG: 4326 "
+                        "is the same as WGS84)."
+                    ),
+                    enum=[DEFAULT_CRS],
+                ),
+            )
+
+        response_headers = [
```
joeribekker, line 66: I'm missing the `API-version` header in OAS responses. Is this actually included in API responses?

joeribekker (changes requested):
I noticed a few differences in the API spec:

* No API-version header in responses
* No error responses (HTTP 401, 403, 404, etc.) included
* Operation ID `object_delete` is now called `object_destroy`
* HTTP responses have no descriptions (`200 OK` vs just `200` -- object_destroy 204 is the exception btw)
* `object_history` seems to have a bunch of query parameters in the OAS that are not actually working? I think they are simply inherited in the docs from the list.
* query parameter `dataAttrs` is now called `data_attrs`?
* `object_search` seems completely wrong. The response is not a list (!), and all query params are gone (`geometry, dataAttrs, date, registrationDate`).

annashamray:
@joeribekker  Thank you for the review!


>     * No API-version header in responses
It was incorrect in the old schema. API headers actually don't exist in the response, because API versioning is turned off for Objects API (for now)
 
>     * No error responses (HTTP 401, 403, 404, etc.) included
I've mentioned it in the description of the PR. The thing is that these error responses in the old schema were incorrect: they had the same shape as in the Openzaak, but error responses in the Objects API don't have this type of body: they are just arbitrary JSONs. If I include them it would not make a lot of sense

>     * query parameter `dataAttrs` is now called `data_attrs`?
it was always called `data_attrs`, it was generated incorrectly in the old schema. 
 
I'll take a look at the rest of the questions and come back

src/objects/utils/autoschema.py
annashamray: No, since api versioning is turned off now. The old schema had it because it used the same generator as Openzaak, where `API-version` header exists

annashamray:
@joeribekker 

> * Operation ID `object_delete` is now called `object_destroy`
fixed

> * HTTP responses have no descriptions (`200 OK` vs just `200` -- object_destroy 204 is the exception btw)
response descriptions are added

> * `object_history` seems to have a bunch of query parameters in the OAS that are not actually working? I think they are simply inherited in the docs from the list.
you are right, I fixed it

> * `object_search` seems completely wrong. The response is not a list (!), and all query params are gone (`geometry, dataAttrs, date, registrationDate`).
I focused on generating schema for `GeometryField` and completely missed that we generate request body from query params here. I've added them, it should be fine now

joeribekker (approved):
Looking good!!

We should bump the version after this to 1.1.1 with all the schema fixes.

PR #205, opened 2021-06-14
Feature/api versioning

### LINKED ISSUE #148: As developer, I want the API to be paginated

### LINKED ISSUE #195: As developer, I want to have access to v1 and v2 of the API

annashamray:
fixes #195 
fixes #148 

depends on #201

### DIFF
--- a/src/objects/tests/v1/utils.py
+++ b/src/objects/tests/v1/utils.py
@@ -1,0 +1,18 @@
+from django.urls import reverse as _reverse
+from django.utils.functional import lazy
+
+VERSION = 1
+
+
+def reverse(viewname, urlconf=None, args=None, kwargs=None, current_app=None):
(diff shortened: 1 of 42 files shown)


src/objects/tests/v1/utils.py
```diff
@@ -0,0 +1,13 @@
+from django.urls import reverse as _reverse
+from django.utils.functional import lazy
+
+VERSION = "v1"
```
joeribekker, line 4: Smart :)

joeribekker (approved):

PR #213, opened 2021-07-01
field based authorization

### LINKED ISSUE #79: As admin-user I want to specify allowed data attributes for object types
- [x] If the user has "write" object permission they can see all attributes, otherwise a list of attributes should be specified
- [x] Only allowed attributes should be returned now  when GET requesting objects
- [x] query param `fields` should be added to explicitly list data attributes 
- [x] If the object has attributes not included in the allowed list, the warning header should be added to the response, which specifies attributes hidden from the user

annashamray:
fixes #79 
This is the first step in implementing field-based authorization

What is left (would be implemented in the next PRs):
* add nice user interface in the admin page to select fields 
* update authorization documentation


src/objects/token/permissions.py
```diff
@@ -43,6 +43,14 @@ def has_object_permission(self, request, view, obj):
         if not object_permission:
             return False
 
+        # forbid showing history if there is field based auth
```
joeribekker, line 46: I think this is a good call for now.

src/objects/tests/v2/test_auth_fields.py
```diff
@@ -0,0 +1,375 @@
...
+            fields=["url", "type", "some"],
+        )
+        object = ObjectFactory.create(object_type=self.object_type)
+        ObjectRecordFactory.create(object=object)
+        url = reverse("object-detail", args=[object.uuid])
+
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(
```
joeribekker, line 109: Good call!

This is probably the best behaviour and should be resolved by the admin.

src/objects/utils/serializers.py
```diff
@@ -1,69 +1,125 @@
...
 
+        self.not_allowed = NotAllowedDict(set)
+
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+
+        allowed_fields = self.get_allowed_fields(instance)
         query_fields = self.get_query_fields()
+
+        if not query_fields and allowed_fields == ["*"]:
```
joeribekker, line 66: Make "*" a constant somewhere to prevent magic strings.

joeribekker (changes requested):

src/objects/api/v1/views.py
```diff
@@ -122,3 +123,20 @@ def get_search_output(self, queryset: models.QuerySet) -> Response:
 
     # for OAS generation
     search.is_search_action = True
+
```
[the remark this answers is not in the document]
annashamray: We don't have different authorization models for v1 and v2, therefore if the administrator turns on field based authorization for some objecttype, API v1 will also be affected. So I think we should also include the header with unauthorized fields here

src/objects/token/permissions.py
annashamray: "history" andpoint has other structure of the response than other endpoints, so I thought this would be the easiest to support solution

src/objects/utils/serializers.py
annashamray: done

joeribekker (approved):

PR #228, opened 2021-07-26
Feature/oas paths

### LINKED ISSUE #226: /api/v1 in endpoint paths
In the objects and objecttypes APIs, the `/api/v1` is always included in the paths part of the schema, while in other APIs like zaken, the `zaken/api/v1` part is included in the base URL (so the individual paths of all the endpoints does not include it). 

It might be good to change it in the objects/objecttypes so that it is consistent with the other APIs.

steps:
* [x] Objects API
* [x] Objecttypes API

annashamray:
fixes #226

### DIFF
--- a/src/objects/api/v1/openapi.yaml
+++ b/src/objects/api/v1/openapi.yaml
@@ -877,6 +877,8 @@ components:
       in: header
       name: Authorization
       description: Token-based authentication with required prefix "Token"
+tags:
+- name: Objects
 externalDocs:
   url: https://objects-and-objecttypes-api.readthedocs.io/
 servers:
--- a/src/objects/api/v1/openapi.yaml
+++ b/src/objects/api/v1/openapi.yaml
@@ -68,7 +68,7 @@ info:
   license:
     name: EUPL-1.2
 paths:
-  /api/v1/objects:
+  /objects:
     get:
       operationId: object_list
       description: Retrieve a list of OBJECTs and their actual RECORD. The actual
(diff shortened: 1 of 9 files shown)


src/objects/api/v1/openapi.yaml
```diff
@@ -92,7 +92,7 @@ paths:
           Data filtering expressions are comma-separated and are structured as follows:
           A valid parameter value has the form `key__operator__value`.
           `key` is the attribute name, `operator` is the comparison operator to be used and `value` is the attribute value.
-          Note: Values can be string or numeric. Dates are not supported.
+          Note: Values can be string, numeric, or dates (ISO format; YYYY-MM-DD).
```
stevenbal, line 95: Just curious: what caused this change?

SilviaAmAm (approved):
Looks good to me :) Just a side comment: I see that Github Actions puts warnings in the `openapi.yaml` files saying that the tags are not  defined in the global tags. Here it says that this is optional (https://swagger.io/docs/specification/grouping-operations-with-tags/), so I don't know if we want to ignore this :upside_down_face:

src/objects/api/v1/openapi.yaml
annashamray: the schema was not updated in one of the merged PRs (https://github.com/maykinmedia/objects-api/pull/223)

annashamray: the schema was not updated in one of the merged PRs (https://github.com/maykinmedia/objects-api/pull/223)

stevenbal (approved):

annashamray:
@SilviaAmAm I've added global tags and the warning seems to disappear.

> Also, should we make an issue for the same thing in objecttypes API?
I think yes for consistency sake. Could you create an issue please?

PR #220, opened 2021-07-12
Feature/fields in auth admin

annashamray:
fixes #202 

what is left:
* ~~decide whether to include objecttype version in the permissions~~ done in #229 229

Issue #202, opened 2021-06-08
As manager, I want to (optionally) authorize API-keys on fields in the Objects API


joeribekker:
For now, we only cover read-access. If write access is given you can read/write all fields.

* [x] Same behaviour as now, to select objecttypes connected to an API-key. Next to each API-key, the Objecttypes are shown, including a link to a "advanced" page (see below).
* [x] Add a detail (advanced) page for each API-key + Objecttype combination to configure field authorizations.
* [x] There is 1 master checkbox: Use field authorizations. If this is unchecked, field authorizations are ignored!
* [x] Add checkboxes for each "static" attribute in the Objects API (creation_date, data, geometry, etc.)
* [x] Introspect the Objecttype schema and allow to cross checkboxes next to the data fields.
* [x] In the API, only return the fields that are checked in the detail page (if "Use field authorizations" is checked).
* [x] update docs 

I see overlap with #79. We can tackle them when needed.

PR #229, opened 2021-07-28
Feature/fields in auth version

annashamray:
issue #202 
depends on #220 

incorporate object_type version into field-based authorization


src/objects/tests/v1/test_auth_fields.py
```diff
@@ -321,6 +347,37 @@ def test_retrieve_query_fields_not_allowed(self):
...
+            data={"name": "other"},
+            version=1,
+        )
+        url = reverse("object-list")
+
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.json(), [{}, {}])
+        self.assertIn("x-unauthorized-fields", response._headers)
```
joeribekker, line 379: I don't see this header documented in the API spec

joeribekker (changes requested):
Looks great and well tested. Also tested it locally, and works as expected.

Few minor things

src/objects/tests/v1/test_auth_fields.py
annashamray: Thanks for catching it. `x-unauthorized-fields` header is added to the OAS

joeribekker (approved):

Issue #232, opened 2021-08-09
As user I want to login to the admin securely using 2FA


joeribekker:
If enabled and if SSO (if present) is not used.

As dev I want still be able to login without 2FA. So it needs to be optional (setting or admin option).

This is for both Objects and Objecttypes API admins.

PR #233, opened 2021-08-10
implement 2fa for the admin

annashamray:
issue #232


annashamray:
The wrong fork of 2FA lib

PR #239, opened 2021-09-06
add icontains to data_attrs operators

### LINKED ISSUE #235: As a developer, I need the possibility to do a non exact partial search on any of the properties.
This is beneficial for search.

**Update after refinement:**

So there would be support for two options:

* [x] `data_attr` query param will have a new operator `icontains` which filters with `ilike` string values - PR #239
* [x] new query param `data_icontains` will be introduced which allows `ilike` for all string values of  the JSON `data` - PR #242

**OLD:**
~~Clearification~~

~~* Add `_cached_data_values` to object model, to store all values from the `data` field~~
~~* Add `data_contains` to the API search endpoint, to search through `_cached_data_values` field~~

annashamray:
fixes #235 

**Upd.** 

Quick performance test which shows that the performance of `data_attrs` query param with `icontains` is in line with others:
```
 Name                                                          # reqs      # fails  |     Avg     Min     Max  Median  |   req/s failures/s
--------------------------------------------------------------------------------------------------------------------------------------------
 GET Retrieve all objects                                         215     0(0.00%)  |     175     113     373     140  |    0.72    0.00
 *GET Retrieve by data_attrs with icontains                       212     0(0.00%)  |      96      73     171      93  |    0.71    0.00
 GET Retrieve by data_attrs with lte                              193     0(0.00%)  |     110      87     206     100  |    0.64    0.00
 GET Retrieve by data_icontains                                   214     0(0.00%)  |     100      79     209      95  |    0.71    0.00
 GET Retrieve by geo coordinates                                  195     0(0.00%)  |     168     114     361     140  |    0.65    0.00
 GET Retrieve single object                                       184     0(0.00%)  |      68      54     180      66  |    0.61    0.00
--------------------------------------------------------------------------------------------------------------------------------------------
 Aggregated                                                      1213     0(0.00%)  |     120      54     373     100  |    4.04    0.00

### DIFF
--- a/src/objects/api/constants.py
+++ b/src/objects/api/constants.py
@@ -9,3 +9,4 @@ class Operators(DjangoChoices):
     gte = ChoiceItem("gte", _("greater than or equal to"))
     lt = ChoiceItem("lt", _("lower than"))
     lte = ChoiceItem("lte", _("lower than or equal to"))
+    icontains = ChoiceItem("icontains", _("case-insensitive containment"))
--- a/src/objects/api/constants.py
+++ b/src/objects/api/constants.py
@@ -9,4 +9,6 @@ class Operators(DjangoChoices):
     gte = ChoiceItem("gte", _("greater than or equal to"))
     lt = ChoiceItem("lt", _("lower than"))
     lte = ChoiceItem("lte", _("lower than or equal to"))
-    icontains = ChoiceItem("icontains", _("case-insensitive containment"))
+    icontains = ChoiceItem(
+        "icontains", _("case-insensitive containment, can be used for strings.")
+    )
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by annashamray** [comment]
this implementations is not what was needed

File: src/objects/api/constants.py [hunk 1]
```diff
@@ -9,3 +9,6 @@ class Operators(DjangoChoices):
     gte = ChoiceItem("gte", _("greater than or equal to"))
     lt = ChoiceItem("lt", _("lower than"))
     lte = ChoiceItem("lte", _("lower than or equal to"))
+    icontains = ChoiceItem(
+        "icontains", _("case-insensitive containment, can be used for strings.")
```
joeribekker, line 13:
```suggestion
        "icontains", _("case-insensitive partial matches on string values.")
```

joeribekker (changes requested):
Looks good but please change the textual description.

joeribekker (approved):

PR #242, opened 2021-09-13
add data_contains search

annashamray:


src/objects/api/filters.py
```diff
@@ -127,6 +120,12 @@ class ObjectFilterSet(FilterSet):
         )
         % {"operator_choices": display_choice_values_for_help_text(Operators)},
     )
+    data_icontains = filters.CharFilter(
+        method="filter_data_icontains",
+        help_text=_(
+            "Search in `data` values when the property name is unknown, can be used fro strings"
```
joeribekker, line 126:
```suggestion
            "Search in all `data` values of string properties."
```

src/objects/utils/views.py
```diff
@@ -23,3 +28,16 @@ def server_error(request, template_name=ERROR_500_TEMPLATE_NAME):
...
     return http.HttpResponseServerError(template.render(context))
+
+
+def exception_handler(exc, context):
+    response = drf_exception_handler(exc, context)
+
+    # provide user-friendly response if data_icontains was used but DB couldn't process it
+    if not response and isinstance(exc, DatabaseError) and "jsonpath" in exc.args[0]:
+        data = {
+            "detail": "Objects DB doesn't support 'data_icontains'. Please, update it to Postgres 12."
```
joeribekker, line 39: This is an error to the user, so the update to Postgres 12 message doesn't make much sense.

```suggestion
            "detail": "This search operation is not supported by the underlying data store."
```

I'm not sure if this 500 is then picked up by Sentry. If not, we should add `logger.exception`.

joeribekker (changes requested):
Looking good, minor remarks.

src/objects/utils/views.py
annashamray: `logger.exception` is added

joeribekker (approved):

Issue #246, opened 2021-09-21
As an admin I want Objects/Objecttypes API to support to OpenID Connect for authentication


stevenbal:
... so I can login to the Objects/Objecttypes admin via Keycloak (or other OIDC providers)

PR #247, opened 2021-09-21
Feature/OIDC integration

stevenbal:
fixes: https://github.com/maykinmedia/objects-api/issues/246

TODO:
- [x] ~~Doesn't seem to work with 2FA (https://github.com/open-formulieren/open-forms/issues/693)~~ -> added setting to disable 2FA in https://github.com/maykinmedia/objects-api/pull/251

### DIFF
--- a/docs/installation/index.rst
+++ b/docs/installation/index.rst
@@ -17,3 +17,4 @@ this.
    quickstart
    config
    deployment/index
+   oidc
--- a/docs/installation/oidc.rst
+++ b/docs/installation/oidc.rst
@@ -1,0 +1,70 @@
+.. _installation_oidc:
+
+============================
+OpenID Connect configuration
+============================
+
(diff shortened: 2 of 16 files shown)


annashamray (changes requested):
Nice! But I think it's better to create staff users with no permissions from the beginning. 

When user tries to log in with OIDC they receive this message:
"You are authenticated as <user id>, but are not authorized to access this page. Would you like to login to a different account? "
And the user can't check if they can actually log in, since we don't have UI except for the admin page. 

Allowing them to enter the empty admin is more user-friendly and at least shows them that their login works 

But perhaps @joeribekker has another option on this.

stevenbal:
> And the user can't check if they can actually log in, since we don't have UI except for the admin page.

If you check the option `make_users_staff` in the OIDC config in the admin, this should be solved I think

annashamray (approved):

annashamray:
Discussed with Joeri. It's fine for now, we can tweak config later

PR #249, opened 2021-09-22
Release/1.2.0 2.0.0

annashamray:


CHANGELOG.rst
```diff
@@ -2,6 +2,48 @@
 Change history
 ==============
 
+2.0.0 (2021-09-22)
```
joeribekker, line 5: Regarding versioning: Let's just use (only) the highest API version as component version.

This means version 2.0.0 of the Objects API supports Objects API specification 1.2.0 and 2.0.0.
The changelog item can therefore be merged, with the mention of the support for both API spec versions.

Also, let's explicetly, mentioned API version changes below this version, so:

2.0.0 (2021-09-22)
------------------

**Component changes**

* Supports API 2.0.0 and API 1.2.0

**API 1.2.0 changes**

* **Breaking features**
  * paginated API responses (#148)
  * ...

**API 2.0.0 changes**

...

joeribekker (changes requested):

CHANGELOG.rst
```diff
@@ -2,6 +2,79 @@
...
+* added superuser for quick-start (#203)
+* tested the performance of the API per version (#219)
+
+**Documentation**
+
+* marked read-only fields as non-required in OAS (#210)
+* described how to configure authorization with the set of allowed fields in the admin (#79)
+* documented how to configure notifications (#245)
+
+
```
joeribekker, line 77: This can be removed then no?

joeribekker (changes requested):

CHANGELOG.rst
annashamray: oops, now it's removed

joeribekker (approved):

PR #254, opened 2021-09-27
:sparkles: Add management command to create initial superuser

stevenbal:
- Adds the `createinitialsuperuser` command (with the option to supply a password instead of generating one)
- Runs this command post migration if the envvars `SUPERUSER_NAME` and `SUPERUSER_PASSWORD` are specified


stevenbal:
closed in favor of https://github.com/maykinmedia/objects-api/pull/255

Issue #257, opened 2021-10-14
Objects API - Searching on integer/numerical values via objects_list's data_attrs __icontains doesn't match objects


alextreme:
Via our friends at Ritense, we have the following object:

                    "bag": {

                        "adres": {

                            "plaats": "AMSTERDAM",

                            "straat": "",

                            "postcode": "1018AK”,

                            "huisnummer": “14”,

                            "toevoeging": "3"

                        }

                    }

You would expect to find this object via:

 /api/v2/objects?data_attrs=bag__adres__huisnummer__icontains__14 

However this doesn't work, the object isn't returned. They have confirmed that using huisnummer__exact__14 does work as intended, the issue is limited to the working of the icontains operator.

I suspect it has to do with '14' in the query being seen as a numeric value instead of a string, causing the matching to mis when comparing to "14". 

The same can be seen when doing a partial-match against the postcode field. They gave the example of postcode__icontains__1018 which doesn't match the object. However, when looking at the letters it does work: postcode__icontains__AK

I suspect this might get tricky, always casting the value to a string when performing the icontains may work in this situation but this may cause issues when doing an icontains on a numeric field?

alextreme:
Input from Erik from Ritense:

The problem is in the string_to_value function on line 93 of v2/filters.py:

```
def filter_data_attrs(self, queryset, name, value: str):
    parts = value.split(",")

    for value_part in parts:
        variable, operator, val = value_part.rsplit("__", 2)
        real_value = string_to_value(val)
```
The following converts a number to a float:
```
def string_to_value(value: str) -> Any:
    if is_number(value):
        return float(value)
    elif is_date(value):
        return datetime.strptime(value, "%Y-%m-%d")

    return value
```
This means that a number like 1011 is converted to 1011.0 in the query. 

It is not clear if this is intentional or not, but removing the conversion to a float leads to the problem being fixed.

annashamray:
@alextreme `icontains` filter supports only string properties now, it's included in the OAS: 
> `icontains` - case-insensitive partial matches on string values.

So I removed `bug` tag from the user story. 

I remember that we haven't included support for numeric values because it's not clear what it the expected behavior of `icontains` on integers? Should it behave exactly like `exact` or should it treat the integer like a string? For example, if I search for `3` should `30` be also in the search results?

PR #261, opened 2021-11-09
Fix for #257 Searching on integer/numerical values via objects_list's…

erik-ritense:
… data_attrs __icontains doesn't match objects

preventing numbers to be written with a decimal point for filter operations such as icontains

### DIFF
--- a/src/objects/api/utils.py
+++ b/src/objects/api/utils.py
@@ -5,9 +5,7 @@ from djchoices import DjangoChoices
 
 
 def string_to_value(value: str) -> Any:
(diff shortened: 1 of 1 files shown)


annashamray:
Hi Erik! Thank you for your interest in improving Objects API. 
I can't approve your PR for two reasons: 
* it breaks filtering on operators used for numeric properties like `lt`, `gt`, `exact`
* it doesn't include regression tests

The required change is being developed here https://github.com/maykinmedia/objects-api/pull/262

Issue #263, opened 2021-11-22
As user I want to know if I should expect objects with geographic coordinates 


annashamray:
The original issue  - https://github.com/GemeenteUtrecht/ZGW/issues/1034

>**as a functional beheerder i would like to use define a variable hasGeometry in an objecttype.**
Context is the Objecttypes API.
This variable can be used to determine how an object should be shown in a frontend. A checkbox can be used to indicate if the objecttype has geometry

PR #265, opened 2021-11-26
Feature/hasgeometry check

annashamray:
issue - #263
explanation - https://github.com/maykinmedia/objecttypes-api/pull/72#pullrequestreview-816096692


src/objects/api/validators.py
```diff
@@ -88,3 +89,35 @@ def validate_data_attrs(value: str):
...
+        self.instance = getattr(serializer, "instance", None)
+
+    def __call__(self, attrs):
+        object_type = attrs.get("object_type") or self.instance.object_type
+        geometry = attrs.get("record", {}).get("geometry")
+
+        if not geometry:
+            return
+
+        client = object_type.service.build_client()
```
joeribekker, line 112: Hmm, this implementation causes 2 calls to Objecttype API, one for the objecttype and one for the geo. Seems inefficient? I can make a new ticket to optimize this but this should not be released as is.

src/objects/api/validators.py
```diff
@@ -88,3 +89,35 @@ def validate_data_attrs(value: str):
...
+        try:
+            response = client.retrieve("objecttype", url=object_type.url)
+        except ClientError as exc:
+
+            msg = f"Object type can not be retrieved: {exc.args[0]}"
+            raise ValidationError(msg)
+
+        has_geometry = response.get("hasGeometry")
```
joeribekker, line 120: What happens if the objecttype API is an older version and this property is not there yet? The default should be to allow it because it was always allowed before.

Please add a test as well.

src/objects/core/models.py
```diff
@@ -122,7 +122,9 @@ class ObjectRecord(models.Model):
         blank=True,
         null=True,
         help_text=_(
-            "Point, linestring or polygon object which represents the coordinates of the object"
+            "Point, linestring or polygon object which represents the coordinates of the "
+            "object. Geometry can be added only if the related OBJECTTYPE allows this "
+            "(`OBJECTTYPE.hasGeometry = true`)"
```
joeribekker, line 127: Or doesn't exist...

joeribekker (changes requested):

src/objects/api/validators.py
annashamray: Actually I don't know how to reduce the number of calls, because `json_schema` validation requests Objecttype.version and `geometry` validation requests Objecttype itself. 
The only solution is to place `hasGeometry` into Objecttype.version but I don't think it's a good decision

src/objects/core/models.py
annashamray: Do you want me to change back the description?

src/objects/api/validators.py
annashamray: Good catch! I made the default value `True` and added a test

src/objects/core/models.py
annashamray: the description is updated

joeribekker (approved):

PR #275, opened 2021-12-23
add ordering to API

annashamray:


annashamray:
ordering on prefetched fields doesn't seem possible, therefore requires refactor of the view, which is done in https://github.com/maykinmedia/objects-api/pull/276

PR #284, opened 2022-03-09
:arrow_up: Bump package versions because of CVEs

damm89:


annashamray:
Sorry I created a new PR for this change (https://github.com/maykinmedia/objects-api/pull/285), since we usually update dependencies without changing /requirements/base.in

PR #286, opened 2022-03-16
Ticket 282 updating starting at value results in 500

### LINKED ISSUE #282: Updating an Object that was previously updated with startAt before initial startAt causes 500 error
Issue raised by Den Haag

Steps to reproduce:
1) Create an Object with some date for `startAt`
2) Update this Object with a date before the `startAt` in step 1
3) Attempt to update this Object again, resulting in a 500 error:
```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "core_objectrecord_correct_id_key"
DETAIL:  Key (correct_id)=(2) already exists.
```
@annashamray I'm not sure what the correct behaviour would be here, should it be possible to update an `Object` and set a `startAt` before the original `startAt`?

bart-maykin:
Fixes #282

Added a test case to recreate the initial problem described on issue 282

1. Create an Object with some date for startAt
2. Update this Object with a date before the startAt in step 1
3. Attempt to update this Object again, resulting in a 500 error:
`django.db.utils.IntegrityError: duplicate key value violates unique constraint "core_objectrecord_correct_id_key"
DETAIL:  Key (correct_id)=(2) already exists.`

After debugging discovered that the issue was the last_record function of the Object model.
Where we ordered on startAt and -index.

The solution to this problem was to remove the startAt from the orderBy.

### DIFF
--- a/src/objects/core/models.py
+++ b/src/objects/core/models.py
@@ -69,7 +69,7 @@ class Object(models.Model):
 
     @property
     def last_record(self):
-        return self.records.order_by("-start_at", "-index").first()
+        return self.records.order_by("-index").first()
 
     @property
     def record(self):
--- a/src/objects/tests/v2/test_object_api.py
+++ b/src/objects/tests/v2/test_object_api.py
@@ -1,4 +1,5 @@
 import json
+import uuid
 from datetime import date, timedelta
 
 import requests_mock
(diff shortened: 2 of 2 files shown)


joeribekker:
Hi @bart-maykin

Without looking at the issue itself, or your fix, please adhere to our best practices.

* Put "Fixes #282" in the PR description as the first line.
* Squash all commits (since its small) into 1
* Change commit message to inclide the ticket number, prefixed by a Gitmoji, like: ":bug: Fixes #282 -- Changed sorting order to ...."

(I see I need to poke @annashamray about this as well)

annashamray (approved):
It's a nice fix of `record.index` generation

PR #307, opened 2022-08-15
[#306] Add `typeVersion` filter to `ObjectRecordQueryset`

### LINKED ISSUE #306: Filter `typeVersion` field on the `/objects` endpoint
Although it doesn't really make sense to filter objects according to their type version, as objects with different objecttypes can have the same version number, it is useful on my end when filtering objects with the `type` url field to have the ability to filter the versions as well.

This shouldn't be too hard to implement, so if you are ok with this feature I can make a PR. (I understand this is something you would not want to have for the reason stated above).

Viicos:
fixes #306

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,7 +65,7 @@ given and if everything is approved, it will be merged.
 
 ### Reviews on releases
 
-All pull requests will be reviewed by a project memeber before they are merged 
+All pull requests will be reviewed by a project member before they are merged 
 to a release branch. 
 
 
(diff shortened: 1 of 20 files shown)


annashamray (changes requested):
Please add tests for `typeVersion` filter

annashamray (changes requested):
Sorry I forgot to mention it in the first review, but please update OAS file. since there is an API change here. 
You can do it with the command `bin/generate_schema.sh v2`

annashamray (approved):

PR #308, opened 2022-08-15
:bug: Fix date parsing in API filtering

Viicos:
Using a `datetime` object instead of the `date` one can result to a wrong filtering behavior in some cases.

Suppose we have a few `ObjectRecord` instances, with a date field in their data. Trying to filter on this field with a `datetime` object and a `gte`/`lte` lookup will result in the following:

 ```python
>>> ObjectRecord.objects.filter(
    data__myDateField__gte=datetime.strptime("2000-02-20", "%Y-%m-%d")
).order_by("data__myDateField").first().data["myDateField"]

'2001-01-10"  # Not the correct object
```

Using a `date` object instead:

```python
>>> ObjectRecord.objects.filter(
    data__myDateField__gte=date.fromisoformat("2000-02-20")
).order_by("data__myDateField").first().data["myDateField"]

'2000-02-00'  # gte lookup works as excepted
```

### DIFF
--- a/src/objects/api/filter_backends.py
+++ b/src/objects/api/filter_backends.py
@@ -22,7 +22,7 @@ class OrderingBackend(OrderingFilter):
     )
 
     def get_valid_fields(self, queryset, view, context={}):
-        """ add nested fields to available fields for ordering"""
+        """add nested fields to available fields for ordering"""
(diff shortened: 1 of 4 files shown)


annashamray (commented):
Could you add a regression test for it?

annashamray (approved):
You can merge if after fixing CI checks

Issue #310, opened 2022-10-28
Add environment settings to Sentry to merge Sentry projects under a different environment


joeribekker:
The Sentry settings in: https://github.com/maykinmedia/objects-api/blob/master/src/objects/conf/base.py#L384:L387

Needs to include more settings, like environment: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/conf/base.py#L732:L736

Same for Objecttypes API.

joeribekker:
@annashamray I think this was fixed right?

PR #312, opened 2022-11-22
Environment config for Sentry

annashamray:
issue - #310

### DIFF
--- a/docs/installation/config.rst
+++ b/docs/installation/config.rst
@@ -57,9 +57,12 @@ can be configured by setting the following environment variables
 
 Other settings
 --------------
+* ``ENVIRONMENT``: An identifier for the environment, displayed in the admin depending on
+  the settings module used and included in the error monitoring (see ``SENTRY_DSN``).
+  The default is set according to ``DJANGO_SETTINGS_MODULE``. Good examples values are:
 
-* ``ADMINS``: Comma seperated list (without spaces!) of e-mail addresses to
-  sent an email in the case of any errors. Defaults to an empty list.
+  * ``production``
+  * ``test``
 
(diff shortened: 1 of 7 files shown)


docs/installation/config.rst
```diff
@@ -57,9 +57,12 @@ can be configured by setting the following environment variables
 
 Other settings
 --------------
+* ``ENVIRONMENT``: An identifier for the environment, displayed in the admin depending on
+  the settings module used and included in the error monitoring (see ``SENTRY_DSN``).
+  The default is set according to ``DJANGO_SETTINGS_MODULE``. Good examples values are:
 
-* ``ADMINS``: Comma seperated list (without spaces!) of e-mail addresses to
```
annashamray, line 61: I removed `ADMINS` env var, it doesn't look like we use it

src/objects/conf/utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Pass ``split=True`` to split the comma-separated input into a list.
+    """
+    if "split" in kwargs:
+        kwargs.pop("split")
+        kwargs["cast"] = Csv()
+
+    if default is not undefined and default is not None:
+        kwargs.setdefault("cast", type(default))
+    return _config(option, default=default, *args, **kwargs)
```
joeribekker, line 24: See openforms latest version for correct code :)

and the test!!!

joeribekker (changes requested):

src/objects/conf/utils.py
annashamray: done

joeribekker (approved):

Issue #324, opened 2023-03-22
Search still returns previous versions of objects for a while after they have been updated


ferdi-ritense:
We ran into an issue where, after we change a property, the previous version of the object still gets returned when searching for the old value of the property.

For example, we have an object with a nested property straat.

```
{
    "adres": {
        "straat": "Bospad"
    }
}
```

The object gets returned, as expected, with this query: `data_attrs=adres__straat__exact__Bospad`

Now we change the straat to a new value. Version 2 of the object now exists with this new value.

```
{
    "adres": {
        "straat": "Dorpsstraat"
    }
}
```
When we search `data_attrs=adres__straat__exact__Dorpsstraat` it behaves as expected. Version 2 of the object is returned.

However, when we search for the old value, `data_attrs=adres__straat__exact__Bospad`, version 1 is returned. But only for a few hours. When we try the same query the next day `data_attrs=adres__straat__exact__Bospad` no longer yields any results which is what we would have expected from the start.

There seems to be some form of caching going on somewhere in the Objects Api. We have reproduced this on a number of environments running versions 2.1.0 (v2) and 2.1.1 (v2). There is no caching outside of the Objects Api in these environments.

ferdi-ritense:
In this example we modified the object from the Objects Api Admin interface.

alextreme:
Investgated further after a hunch of Ivo, the PR seems to solve the problem (Joeri is having a look)

joeribekker:
Made a deep dive into this. The "problem" here is the data filter.

The regular order of things goes like this:

1. Filter records (if no date provided, it searches on valid records for today)
2. If multiple records match, do conflict resolution.

Assuming we have an `Object O` with 2 records: `Record A` with `prop=K` and a newer `record B` with `prop=L`. In essence, we have an object with some property with value `K` and we updated it to say the property is now `L`.
`Record A` started on `t1` and ended on `t2`. At date `t2`, `record B` starts (without any end date).

```
Object O records:
|-- A(prop=K) --|
                |-- B(prop=L) --
t1              t2
```

Which record is valid on date `t2`? In this case that would be `record B`.

* Rule 1: Both records match 
* Rule 2: `Record B` is newer and therefore wins the "conflict"
* Result: `Record B` **correct**

*If you ask why did `record A` match at all? Well, imagine that record B didn't exist. Should `record A` be returned then when search in on date `t2`? Yes it should, so that's your answer.*

[...]

PR #325, opened 2023-03-23
[#324] Having startAt and endAt on today caused former objects to be shown

alextreme:
Resolves the same-date issue when retrieving objects


alextreme:
Resolved failing 'ci' coordinate-tests by pinning testrunner to ubuntu-20.04 (ubuntu-latest switched to 22.04 somewhere last year).

quick-start should be working again after #327 is merged into master (as this action always runs against master for some reason).

src/objects/core/query.py
```diff
@@ -16,7 +16,7 @@ def filter_for_date(self, date):
         return (
             self.filter(records__start_at__lte=date)
             .filter(
-                models.Q(records__end_at__gte=date)
+                models.Q(records__end_at__gt=date)
```
joeribekker, line 19: I gave this some thought and my main concern was that this changed the filtering and not the data. Should the previous record not get an end date that is start date (of new record) - 1?

If the last record of an object, ends today. Is it valid today? According to the data it is but according to the filtering done here it doesn't match.

The docs are unclear what endsAt means (inclusive or exclusive). I need to ponder in this some more.

alextreme: As discussed earlier today I'd read 'endsAt=29-03-2023' as that the object ends at 29-03-2023 00:00 (if the date is 29-03-2023 it's validity has ended, not that 29-03-2023 is the last day that the object still is valid). But GPT-4 disagreed with me:

> Given the API specification, it is not explicitly mentioned whether the endAt date should be considered inclusive or exclusive. However, since the API deals with history and time periods, it would be reasonable to assume that the endAt date is inclusive. This means the object would still be considered valid on the 'endsAt' date.

It is indeed not explicit, but the alternative is that creating a new object and updating it on the same day without an explicit endAt means that multiple versions of the object will be valid on the current day. So the alternative (updating the object and updating the previous record's endAt to yesterday? I'm not sure that's possible via the API) seems less logical ( the API-test added shows the expected behaviour, I'm not sure how you would want to do this otherwise: https://github.com/maykinmedia/objects-api/pull/325/files#diff-c900604ca49840241e6d00d829c50ad83d98cbe15ce867417ffeac500322aba9 ).

Using datetimes instead of dates would make this explicit and the most sense in my eyes, I don't see why an object couldn't be valid for only a few seconds instead of a full day, but that's a different discussion.

joeribekker: In thought I explained another approach.

I'll write some code to demonstrate.

joeribekker:
I made a tag for this to resolve this temporarily with Alex' fix: `2.1.1-rt324`

PR #338, opened 2023-06-06
⬆️  Upgrade pip packages

ErhanCitil:
Patch Tuesday,

When you try to install the requirements from base.txt it gives you an error. This is because an older version of the packages: **uwsgi & ruamel yaml clib**

In this pull request I updated the dependencies to the latest version, so it doesn't give you an error when installing.


alextreme (approved):
Approved as it's Patch Tuesday, hope you don't mind @annashamray

Issue #351, opened 2023-12-01
Partial update (PATCH) of an object replaces data


SilviaAmAm:
After creating an object with multiple properties in the "data" field, doing a PATCH call to update one of the properties in data overwrites the whole data field.

# Steps to reproduce

(with a fake URL for the environment, but tested on our test environment)

1. Create an object: POST https://objecten.test.nl/api/v2/objects

```json
{
    "type": "https://objecttypen.test.nl/api/v2/objecttypes/2360a778-6747-47b4-ad2b-e5e281336bba",
    "record": {
        "typeVersion": 2,
        "data": {
            "type": "Fruitteelt",
            "leeftijd": 100,
            "kiemjaar": 2002
        },
        "geometry": {
            "type": "Point",
            "coordinates": [
                4.766006469726562,
                52.29066294826172
            ]
        },
        "startAt": "2023-12-01"
    }
}
```
2. Update the object (PATCH https://objecten.test.nl/api/v2/objects/19fadf7f-5695-41a1-9de8-a7246b1c28a3)

```json
{
    "record": {
        "data": {
            "kiemjaar": 2012
        },
        "startAt": "2023-12-01"
    }
}
```

This then returns the following object:
```json
{
    "url": "https://objecten.test.nl/api/v2/objects/19fadf7f-5695-41a1-9de8-a7246b1c28a3",
    "uuid": "19fadf7f-5695-41a1-9de8-a7246b1c28a3",
    "type": "https://objecttypen.test.nl/api/v2/objecttypes/2360a778-6747-47b4-ad2b-e5e281336bba",
    "record": {
        "index": 2,
        "typeVersion": 2,
        "data": {
            "kiemjaar": 2012
        },
        "geometry": null,
        "startAt": "2023-12-01",
        "endAt": null,
        "registrationAt": "2023-12-01",
        "correctionFor": null,
        "correctedBy": null
    }
}
```
The `data` field is missing the `type` and the `leeftijd` attributes that were set when first creating the object.

joeribekker:
Refinement: 

Discussion involves: 

Add support for `Content-Type: application/merge-patch+json` to merge the "data"

* https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/
* https://datatracker.ietf.org/doc/html/rfc7386

PR #352, opened 2023-12-12
✨ Use JSON Merge Patch when doing a partial update on records

Viicos:
Fixes https://github.com/maykinmedia/objects-api/issues/351

### DIFF
--- a/src/objects/api/serializers.py
+++ b/src/objects/api/serializers.py
@@ -10,6 +10,7 @@ from objects.utils.serializers import DynamicFieldsMixin
 
 from .fields import ObjectSlugRelatedField, ObjectTypeField, ObjectUrlField
 from .validators import GeometryValidator, IsImmutableValidator, JsonSchemaValidator
+from .utils import merge_patch
 
 
 class ObjectRecordSerializer(serializers.ModelSerializer):
@@ -129,6 +130,11 @@ class ObjectSerializer(DynamicFieldsMixin, serializers.HyperlinkedModelSerialize
         # in case of PATCH
         if "version" not in validated_data:
             validated_data["version"] = instance.version
+            if "data" in validated_data:
+                # Apply JSON Merge Patch for record data
+                validated_data["data"] = merge_patch(
+                    instance.data, validated_data["data"]
+                )
 
         record = super().create(validated_data)
         return record
--- a/src/objects/api/serializers.py
+++ b/src/objects/api/serializers.py
@@ -9,8 +9,8 @@ from objects.token.models import Permission
 from objects.utils.serializers import DynamicFieldsMixin
 
 from .fields import ObjectSlugRelatedField, ObjectTypeField, ObjectUrlField
-from .validators import GeometryValidator, IsImmutableValidator, JsonSchemaValidator
 from .utils import merge_patch
+from .validators import GeometryValidator, IsImmutableValidator, JsonSchemaValidator
 
 
 class ObjectRecordSerializer(serializers.ModelSerializer):
--- a/src/objects/api/serializers.py
+++ b/src/objects/api/serializers.py
@@ -127,14 +127,12 @@ class ObjectSerializer(DynamicFieldsMixin, serializers.HyperlinkedModelSerialize
         # object_data is not used since all object attributes are immutable
         object_data = validated_data.pop("object", None)
         validated_data["object"] = instance.object
-        # in case of PATCH
+        # version should be set
         if "version" not in validated_data:
             validated_data["version"] = instance.version
-            if "data" in validated_data:
-                # Apply JSON Merge Patch for record data
-                validated_data["data"] = merge_patch(
-                    instance.data, validated_data["data"]
-                )
+        if self.partial and "data" in validated_data:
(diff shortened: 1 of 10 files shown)


src/objects/api/utils.py
```diff
@@ -43,3 +43,22 @@ def display_choice_values_for_help_text(choices: DjangoChoices) -> str:
         items.append(item)
 
     return "\n".join(items)
+
+
+def merge_patch(target: Any, patch: Any) -> Dict[str, Any]:
```
SilviaAmAm, line 48: target and patch should always be JSON values, so maybe the type hints can be more specific?

src/objects/api/utils.py
```diff
@@ -43,3 +43,22 @@ def display_choice_values_for_help_text(choices: DjangoChoices) -> str:
...
+        return patch
+
+    if not isinstance(target, dict):
+        # Ignore the contents and set it to an empty dict
+        target = {}
+    for k, v in patch.items():
+        if v is None:
+            if k in target:
+                # remove the key/value pair from target
+                del target[k]
```
SilviaAmAm, line 60: By looking at the tests, I see that this is the desired behaviour, but I wonder about the case where we want to set a key to a null value (for example if the key is required but can take null values) :thinking:

src/objects/api/utils.py
Viicos: This is true for the first call to `merge_patch`, but nested calls doesn't necessarily have dicts as values.

In fact merge Patch can also be applied for any kind of data (you could have arrays for example, even if it doesn't really make sense to patch an array with another one).

src/objects/api/utils.py
Viicos: This is following the RFC implementation, but I agree when writing this I thought about this use case. So we could maybe deviate from the spec, depending on which use case we cant to be able to achieve

src/objects/api/utils.py
SilviaAmAm: What I had in mind is something like what we have in OF, where we define these types:

```python
JSONPrimitive = Union[str, int, None, float, bool]

JSONValue = Union[JSONPrimitive, "JSONObject", List["JSONValue"]]
```
And then I thought maybe instead of `Any` we could use `JSONValue`?

Viicos: Thought about that as well, but for such a small use case decided to go with `Any`. But I think this is still a good idea as typing JSON arguments is common so this could be reused.

src/objects/api/serializers.py
```diff
@@ -129,6 +130,11 @@ def update(self, instance, validated_data):
         # in case of PATCH
         if "version" not in validated_data:
             validated_data["version"] = instance.version
+            if "data" in validated_data:
+                # Apply JSON Merge Patch for record data
+                validated_data["data"] = merge_patch(
```
annashamray, line 135: It should be 
* -1 level of indetation level 
* only for PATCH

src/objects/api/v2/views.py
```diff
@@ -42,7 +42,9 @@
         description="Update the OBJECT by creating a new RECORD with the updates values."
     ),
     partial_update=extend_schema(
-        description="Update the OBJECT by creating a new RECORD with the updates values."
+        description="Update the OBJECT by creating a new RECORD with the updates values. "
+        "The provided `record.data` value will be merged with the existing record data, according "
+        "to [RFC7396](https://datatracker.ietf.org/doc/html/rfc7396)."
```
annashamray, line 47: Please add the same description to v1 view, since its behavior was also changed

src/objects/tests/test_merge_patch.py
```diff
@@ -0,0 +1,29 @@
...
+            ({"a": "c"}, {"a": ["b"]}, {"a": ["b"]}),
+            ({"a": {"b": "c"}}, {"a": {"b": "d", "c": None}}, {"a": {"b": "d"}}),
+            ({"a": [{"b": "c"}]}, {"a": [1]}, {"a": [1]}),
+            (["a", "b"], ["c", "d"], ["c", "d"]),
+            ({"a": "b"}, ["c"], ["c"]),
+            ({"a": "foo"}, None, None),
+            ({"a": "foo"}, "bar", "bar"),
+            ({"e": None}, {"a": 1}, {"e": None, "a": 1}),
+            ([1, 2], {"a": "b", "c": None}, {"a": "b"}),
+            ({}, {"a": {"bb": {"ccc": None}}}, {"a": {"bb": {}}}),
```
annashamray, line 24: Please add a test with the same `target` and `patch`to check idempotency something like
```{"a": "b"}, {"a": "b"}, {"a": "b"}),```

annashamray (changes requested):
I have a general question:

The standard https://datatracker.ietf.org/doc/html/rfc7396 explicitly says that it introduces the new content type `application/merge-patch+json`. In Objects API only `application/json` content type is supported. 
If Objects API starts supporting RFC 7396 behavior, I think it should support its content type. 
@joeribekker what do you think?

annashamray (approved):
Looks very nice and neat!

PR #356, opened 2024-01-17
Update format tools

Viicos:


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Black update
+de05ba4ccf2c9a3c929705cee6f517bf74361a2c
```
annashamray, line 2: TIL about ignoring revisions for blame. :+1:

annashamray (approved):

PR #368, opened 2024-03-05
add `setup_configuration` command

annashamray:


docs/installation/config_cli.rst
```diff
@@ -0,0 +1,135 @@
...
+* ``DEMO_EMAIL``: demo user email. Required.
+* ``DEMO_TOKEN``: demo token. Required.
+
+
+Execution
+=========
+
+
+With the full command invocation, everything is configured at once and immediately
+tested.
```
stevenbal, line 116: I don't think we have a readthedocs for setup-configuration yet, but maybe it would be nice to have a reference to that here at some point

stevenbal (approved):

stevenbal:
Looks good, just one question and a remark

docker-compose.yml
```diff
@@ -30,14 +30,32 @@ services:
...
       - CELERY_LOGLEVEL=DEBUG
+      # setup_configuration env vars
+      - OBJECTS_DOMAIN=web:8000
+      - OBJECTS_ORGANIZATION=Objects
+      - OBJECTTYPES_API_ROOT=https://objecttypes.example.com/api/v2/
+      - OBJECTS_OBJECTTYPES_TOKEN=some-random-string
+      - DEMO_CONFIG_ENABLE=yes
+      - DEMO_TOKEN=demo-random-string
+      - DEMO_PERSON=Demo
+      - DEMO_EMAIL=demo@demo.local
```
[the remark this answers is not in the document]
annashamray: `docker-compose-quickstart.yml` was added with the idea of giving quick demo. So the user can follow steps in README and very quickly have Objects API up and running. And it uses the published docker image, so you can just download this file without cloning the whole repo.

I don't know if it's still needed to be honest. You can ask Joeri, he was behind this quickstart thing

docs/installation/config_cli.rst
annashamray: Good idea. For now setup_configuration lib has just README, I'll put a link for it here

PR #369, opened 2024-03-08
Feature/superuser

annashamray:


docs/admin/authorization.rst
```diff
@@ -88,3 +88,19 @@ fields you can submit the form.
...
 If you want to know how to use Objects API you can follow :ref:`api_usage`
+
+
+Superuser permissions
+----------------------
+
+It's possible to set superuser permissions in Objects API. A client with such permissions
+is able to request objects for all objecttypes.
+
+In the admin page of the Objects API go to the Token authorizations" resource and click on
```
Viicos, line 99:
```suggestion
In the admin page of the Objects API go to the "Token authorizations" resource and click on
```

src/objects/api/fields.py
```diff
@@ -42,7 +44,23 @@ def to_internal_value(self, data):
         try:
             return self.get_queryset().get_by_url(data)
         except ObjectDoesNotExist:
-            self.fail("does_not_exist", value=smart_text(data))
+            # if service is configured, but objec_type is missing
```
Viicos, line 47:
```suggestion
            # if service is configured, but object_type is missing
```

src/objects/api/fields.py
```diff
@@ -42,7 +44,23 @@ def to_internal_value(self, data):
...
+            # let's try to create an ObjectType
+            service = Service.get_service(data)
+            if not service:
+                self.fail("does_not_exist", value=smart_text(data))
```
Viicos, line 51:
```suggestion
                self.fail("does_not_exist", value=smart_str(data))
```

I think when rebasing on the 4.2 upgrade the import will be updated already

docs/admin/authorization.rst
```diff
@@ -88,3 +88,19 @@ fields you can submit the form.
 Now the client who has this token can access the objects with the "Boom" object type.
 
 If you want to know how to use Objects API you can follow :ref:`api_usage`
+
+
+Superuser permissions
+----------------------
+
+It's possible to set superuser permissions in Objects API. A client with such permissions
```
Viicos, line 96:
```suggestion
It's possible to set up superuser permissions in Objects API. A client with such permissions
```

docs/admin/authorization.rst
Viicos: seems to be not applied

src/objects/api/fields.py
```diff
@@ -42,7 +44,23 @@ def to_internal_value(self, data):
...
+            if not service:
+                self.fail("does_not_exist", value=smart_str(data))
+
+            uuid = get_uuid_from_path(data)
+            object_type = self.get_queryset().model(service=service, uuid=uuid)
+
+            try:
+                object_type.clean()
+            except ValidationError:
+                self.fail("does_not_exist", value=smart_str(data))
```
Viicos, line 59: Is the error code a copy paste mistake?

SonnyBA (approved):

src/objects/api/fields.py
annashamray: No. In `object_type.clean()` we try to request the object type url. If it's not existed `ValidationError` will be raised.

Viicos (approved):

PR #376, opened 2024-03-21
:bug: copy celery binary

sjoerdie:
- Celery container crashes because the celery binary does not exist;
- Add `check_celery_worker_liveness.py` , we use it for healthchecks on the celery containers;


annashamray (approved):
Thank you for the fix

PR #380, opened 2024-03-21
[#379] Make use of Python3.11

### LINKED ISSUE #379: Upgrade to Python3.11
Currently the objects API runs on Py3.10, with Py3.11 I get an lxml error.

I'll create a PR to make use of Py3.11 as I think it's an unwritten rule at Maykin to only use odd versions of Python (and to at least make use of the current version in Debian stable)

* [x] Objects API
* [ ] Objecttypes API

alextreme:
fixes #379 (Objects API part)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: '3.11'
       - uses: actions/setup-node@v4
         with:
           node-version: '18'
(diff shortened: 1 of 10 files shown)


annashamray (commented):
Let's discuss in the refinement? I think strange to approve it while Objecttypes still doesn't support Django4.2 yet

Afaik the first prio is to switch to django4.2 for Objecttypes, ON and OZ
And we also started switching to open-api-framework
Should we make some epic with some roadmap to upgrade\patch all APIs ?

alextreme:
Makes sense, this was just because I'm running Py3.11 locally but ideally it should be done together with Django4.2

annashamray:
@alextreme I thing the time has come :)

stevenbal (approved):

Issue #381, opened 2024-03-21
Improve admin search to look within the Objects data


alextreme:
Currently the Objects admin search only looks at the uuid. Which is useful for the narrow situation where you're looking for a single object and you have a uuid.

For regular admin purposes it's more practical to also search through the objects API record data, so I'll create a PR which does just that.

PR #382, opened 2024-03-21
[#381] Improve admin search to also search through object data

alextreme:


annashamray (approved):
I'll approve but does it really work with JSON fields?

alextreme:
Yes, I actually tried this out myself (reason of PR #380 ) because I also couldn't believe it would be this easy.

https://docs.djangoproject.com/en/5.0/topics/db/queries/#querying-jsonfield

PR #393, opened 2024-04-18
:sparkles: [#315] added extra version and uuid fields in admin

### LINKED ISSUE #315: As admin I want to see the UUID and version of an objecttype
Add a column to show the Objecttype UUID ~and version.~:

* [x] List `/admin/core/objecttype/` 
* [x] Inline `/admin/token/tokenauth/{uuid}/change/`
* [x] List /admin/token/permission/`

In the Object admin page:
* [x] Add the Object UUID and the Objecttype UUID columns to the list view
* [x] Add the Objecttype UUID to the detail view

This makes debugging much easier.

PS. In light of recent review, ticket was respecced

bart-maykin:
Fixes #315

**Changes**

Added some version and uuid fields in the admin panel.

### DIFF
--- a/src/objects/core/models.py
+++ b/src/objects/core/models.py
@@ -40,6 +40,17 @@ class ObjectType(models.Model):
         # zds_client.get_operation_url() can be used here but it increases HTTP overhead
         return f"{self.service.api_root}objecttypes/{self.uuid}"
 
+    @property
+    def version(self):
+        client = self.service.build_client()
+        try:
+            object_type_data = client.retrieve("objecttype", url=self.url)
+        except (ClientError, ConnectionError, ValueError) as exc:
+            return None
+
+        if recent_version := object_type_data["versions"][-1]:
+            return recent_version.split("/versions/")[-1]
+
     def clean(self):
         client = self.service.build_client()
         try:
--- a/src/objects/core/models.py
+++ b/src/objects/core/models.py
@@ -40,17 +40,6 @@ class ObjectType(models.Model):
         # zds_client.get_operation_url() can be used here but it increases HTTP overhead
         return f"{self.service.api_root}objecttypes/{self.uuid}"
 
-    @property
-    def version(self):
-        client = self.service.build_client()
-        try:
-            object_type_data = client.retrieve("objecttype", url=self.url)
-        except (ClientError, ConnectionError, ValueError) as exc:
-            return None
-
-        if recent_version := object_type_data["versions"][-1]:
-            return recent_version.split("/versions/")[-1]
-
     def clean(self):
         client = self.service.build_client()
         try:
--- a/src/objects/token/admin.py
+++ b/src/objects/token/admin.py
@@ -14,7 +14,30 @@ EMPTY_FIELD_CHOICE = ("", "---------")
 
 @admin.register(Permission)
 class PermissionAdmin(admin.ModelAdmin):
-    list_display = ("token_auth", "object_type", "mode", "use_fields")
+    list_display = (
+        "token_auth",
+        "object_type",
+        "mode",
+        "use_fields",
+        "get_uuid",
+        "get_version",
+    )
+    readonly_fields = (
+        "get_uuid",
+        "get_version",
+    )
+
+    def get_uuid(self, obj):
+        if obj:
+            return obj.object_type.uuid
+
+    def get_version(self, obj):
+        if obj:
+            if object_type := obj.object_type:
+                return object_type.version
+
+    get_version.short_description = "Version"
+    get_uuid.short_description = "UUID"
 
     def get_object_fields(self):
         object_serializer = ObjectSerializer()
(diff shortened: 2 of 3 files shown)


src/objects/core/models.py
```diff
@@ -40,6 +40,17 @@ def url(self):
         # zds_client.get_operation_url() can be used here but it increases HTTP overhead
         return f"{self.service.api_root}objecttypes/{self.uuid}"
 
+    @property
+    def version(self):
```
annashamray, line 44: It's not a version, it's rather the last available version. But the thing is that we specify the objecttype version in the object data, Objecttype model has only a link to objecttype itself.
@joeribekker Do we really want in the list admin view to perform a lot of HTTP requests to get the last version? 

Even if you want to go this was, I think it's more useful to show the list of all versions available, not the last one. 
And if you want to see them in the list view, it's better to perform one list request to Objecttypes API instead of a lot of individual detail requests

src/objects/token/admin.py
```diff
@@ -104,7 +127,27 @@ def add_view(self, request, form_url="", extra_context=None):
...
     model = Permission
     fk_name = "token_auth"
-    fields = ("object_type", "mode", "use_fields", "fields")
+    fields = (
+        "object_type",
+        "mode",
+        "use_fields",
+        "fields",
+        "get_uuid",
+        "get_version",
```
annashamray, line 136: I think it's confusing, because it looks like the permission works only for this version. In fact any permission works for all objecttype versions. 

@joeribekker The same comment - objecttype versions are not taken into account in permissions, so I don't know if displaying last version would be of any use. 
Do you still want it?

src/objects/token/admin.py
```diff
@@ -104,7 +127,27 @@ def add_view(self, request, form_url="", extra_context=None):
...
+    )
+
+    def get_uuid(self, obj):
+        if obj:
```
annashamray, line 140: Why do you need this condition? It can be removed. This comment is applied for all admin fields

src/objects/token/admin.py
```diff
@@ -104,7 +127,27 @@ def add_view(self, request, form_url="", extra_context=None):
...
+    fields = (
+        "object_type",
+        "mode",
+        "use_fields",
+        "fields",
+        "get_uuid",
+        "get_version",
+    )
+
+    def get_uuid(self, obj):
```
annashamray, line 139:
```suggestion
    def get_objecttype_uuid(self, obj):
```

src/objects/token/admin.py
```diff
@@ -104,7 +127,27 @@ def add_view(self, request, form_url="", extra_context=None):
...
+            if object_type := obj.object_type:
+                return object_type.uuid
+
+    def get_version(self, obj):
+        if obj:
+            if object_type := obj.object_type:
+                return object_type.version
+
+    get_version.short_description = "Version"
+    get_uuid.short_description = "UUID"
```
annashamray, line 150: please use `admin.display` decorator for it. And it's not just uuid, it's "Object type UUID"
Same is applied for versions.

annashamray (changes requested):

src/objects/core/models.py
SonnyBA, line 44: Functionally I don't know how these models interact with each other but in the previous PR from Erhan I saw that [he set the version](https://github.com/maykinmedia/objects-api/pull/332/files#diff-6917b29ef1787115c2db3a76aabd3c9f3eb76dae6796598a38506a140fbaef26R61) (in a new model field) when the `clean` method was called. Would this be a alternative solution (and prevent unnecessary API calls)?

joeribekker: I think I specced this ticket too easily. And we dont want to do many http calls here. I will respec the ticket.

src/objects/token/admin.py
joeribekker: nop, will respecc the ticket.

src/objects/token/admin.py
```diff
@@ -14,7 +14,19 @@
...
+        "object_type",
+        "mode",
+        "use_fields",
+        "get_uuid",
+    )
+    readonly_fields = ("get_uuid",)
+
+    @admin.display(description="Object type UUID")
+    def get_uuid(self, obj):
+        if obj:
```
annashamray, line 28: This condition is not needed

src/objects/token/admin.py
```diff
@@ -104,7 +116,18 @@ def add_view(self, request, form_url="", extra_context=None):
...
+        "object_type",
+        "mode",
+        "use_fields",
+        "fields",
+        "get_uuid",
+    )
+
+    @admin.display(description="Object type UUID")
+    def get_uuid(self, obj):
+        if object_type := obj.object_type:
```
annashamray, line 129: This condition is not needed

src/objects/token/admin.py
```diff
@@ -14,7 +14,19 @@
...
 class PermissionAdmin(admin.ModelAdmin):
-    list_display = ("token_auth", "object_type", "mode", "use_fields")
+    list_display = (
+        "token_auth",
+        "object_type",
+        "mode",
+        "use_fields",
+        "get_uuid",
+    )
+    readonly_fields = ("get_uuid",)
```
annashamray, line 24: We rewrite the template for the token admin page with react, so you need also to add the new field in the `permission-form.js` file

annashamray (changes requested):

src/objects/token/admin.py
bart-maykin: This wasn't asked for in the ticket so lets ignore it for now.

annashamray: In this case please remove `get_uuid` from readonly_fields, because it's not shown in the detailed page

annashamray (changes requested):
Please rebase on master and resolve the conflcts

src/objects/token/admin.py
annashamray: I think this is not yet addressed?

annashamray (commented):
Please remove get_uuid from readonly fields in `PermissionAdmin`. if it's not shown on the detail page,

PR #397, opened 2024-05-02
remove primary key from token

annashamray:
Fixes https://github.com/maykinmedia/security-issues/issues/1 (Objects API part)


joeribekker (approved):
Tested and works.

PR #398, opened 2024-05-02
:green_heart: update demodata with token model change

annashamray:
The goal of this PR to fix CI in master branch and make it green again. 

**Explanation why it is broken:**
* We have quick-start CI workflow which uses Objects API image from docker hub and fixtures inside this image
* In the PR #397 the structure of Token model was changed, but the related fixtures were not updated
* The CI was green in the PR because quick-start action uses docker image from docker hub with old Token model structure
* After the PR was merged the docker image on docker hub was updated, so now we have a docker image with 'latest' tag where demodata fixture and token model have different structures. 
* So to fix it we need to publish a new docker image with updated fixture but to do so we need green CI
* Therefore loading fixtures is temporarily disabled until the fixtures are updated in the docker image on doker hub

### DIFF
--- a/.github/workflows/quick-start.yml
+++ b/.github/workflows/quick-start.yml
@@ -15,8 +15,9 @@ jobs:
         run: docker-compose -f docker-compose-qs.yml up -d
       - name: Wait until DB container starts
         run: sleep 10
-      - name: Load fixtures
-        run: docker-compose -f docker-compose-qs.yml exec -T web src/manage.py loaddata demodata
+#        TODO uncomment when correct fixtures are uploaded into dockerhub inside docker image
+#      - name: Load fixtures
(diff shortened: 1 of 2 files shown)


joeribekker (commented):
> The CI was green in the PR because quick-start action uses docker image from docker hub with old Token model structure

Isn't the proper fix to test against the image that was built in the CI-pipeline instead of the latest Docker image on Dockerhub?

annashamray:
We use demodata fixture only for quick-start, which uses the image published on docker hub. Do we really need this?

joeribekker (approved):

joeribekker:
> We use demodata fixture only for quick-start, which uses the image published on docker hub. Do we really need this?

Ah, you are right. The whole goal of that CI-task is to see if the quickstart steps work, which involves downloading the image. OK, I approved it :)

PR #417, opened 2024-06-12
Issue/373 admin add permission

### LINKED ISSUE #373: 500 error when trying to add a permission and Objecttypes API is unreachable
Accessing `/admin/token/permission/add/` raises a 500 error with the following unhandled exception:

`HTTPConnectionPool(host='objecttypes-host', port=...): Max retries exceeded with url: /api/v2/objecttypes/<uuid>/versions`

annashamray:
Fixes #373

### DIFF
--- a/src/objects/token/admin.py
+++ b/src/objects/token/admin.py
@@ -88,7 +88,7 @@ class PermissionAdmin(admin.ModelAdmin):
             self.get_data_field_choices() if objecttypes_available else {}
         )
 
-        context = {
+        return {
             "object_fields": self.get_object_fields(),
             "data_field_choices": data_field_choices,
             "token_auth_choices": token_auth_choices,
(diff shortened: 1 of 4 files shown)


src/objects/token/admin.py
```diff
@@ -83,20 +83,33 @@ def get_extra_context(self, request, object_id):
...
             "object_fields": self.get_object_fields(),
-            "data_field_choices": self.get_data_field_choices(),
+            "data_field_choices": data_field_choices,
             "token_auth_choices": token_auth_choices,
             "object_type_choices": object_type_choices,
             "mode_choices": mode_choices,
             "form_data": self.get_form_data(request, object_id),
+            "objecttypes_available": objecttypes_available,
         }
+        print(context)
```
stevenbal, line 100: This print can probably be removed

stevenbal (approved):
Minor comment regarding trailing print statement, looks good otherwise :+1:

src/objects/token/admin.py
annashamray: Oops thanks for catching it

PR #421, opened 2024-07-02
:lipstick: fix help-text icon layout in the admin

annashamray:
Fixes https://github.com/open-zaak/open-notificaties/issues/150 (Objects API part)


Coperh (approved):
Yep works fine

PR #425, opened 2024-07-23
 📝 [open-zaak/open-zaak#1649] Document envvars 

stevenbal:
Fixes open-zaak/open-zaak#1649 partially

Requires https://github.com/maykinmedia/open-api-framework/pull/22 to be merged and released

**Changes**

* Use envvar documentation mechanism from OAf
* Add CI job to check if envvar docs were updated

### DIFF
--- a/docs/installation/config.rst
+++ b/docs/installation/config.rst
@@ -1,109 +1,100 @@
-.. _installation_environment_config:
+.. _installation_env_config:
 
 ===================================
 Environment configuration reference
 ===================================
 
-The Objects and Objecttypes APIs can be run both as a Docker container or a VPS / dedicated server. It relies on other services, such as
-database and cache backends, which can be configured through environment
-variables.
+
+The Objects and Objecttypes APIs can be run both as a Docker container or a VPS / dedicated server.
+It relies on other services, such as database and cache backends, which can be configured through environment variables.
+
 
 Available environment variables
 ===============================
 
-Required settings
------------------
 
-* ``DJANGO_SETTINGS_MODULE``: which environment settings to use. Available options:
+Required
+--------
 
-  - ``objects.conf.docker`` or ``objecttypes.conf.docker``
-  - ``objects.conf.dev`` or ``objecttypes.conf.dev``
-  - ``objects.conf.ci`` or ``objecttypes.conf.ci``
+* ``SECRET_KEY``: Secret key that's used for certain cryptographic utilities. You should generate one via `miniwebtool <https://www.miniwebtool.com/django-secret-key-generator>`_.
+* ``ALLOWED_HOSTS``: a comma separated (without spaces!) list of domains that serve the installation. Used to protect against Host header attacks. Defaults to: ``(empty string)``.
 
-* ``SECRET_KEY``: secret key that's used for certain cryptographic utilities. You
-  should generate one via
-  `miniwebtool <https://www.miniwebtool.com/django-secret-key-generator/>`_
 
-* ``ALLOWED_HOSTS``: A comma separated (without spaces!) list of domains that
-  serve the installation. Used to protect against ``Host`` header attacks.
-  Defaults to ``*`` for the ``docker`` environment and defaults to
-  ``127.0.0.1,localhost`` for the ``dev`` environment.
+Database
+--------
 
-Common settings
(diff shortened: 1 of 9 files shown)


docs/installation/config.rst
```diff
@@ -1,109 +1,99 @@
...
-===============================
-
-Required settings
------------------
-
-* ``DJANGO_SETTINGS_MODULE``: which environment settings to use. Available options:
-
-  - ``objects.conf.docker`` or ``objecttypes.conf.docker``
-  - ``objects.conf.dev`` or ``objecttypes.conf.dev``
-  - ``objects.conf.ci`` or ``objecttypes.conf.ci``
```
stevenbal, line 21: This is now removed from the docs, because it is not explicitly set using `config(...)`. Is this issue?

src/objects/conf/base.py
```diff
@@ -72,13 +72,24 @@
...
+    default=15 * 60,
+    help_text=(
+        "Task hard time limit in seconds. The worker processing the task will be "
+        "killed and replaced with a new one when this is exceeded."
+    ),
+    group="Celery",
+)  # hard
+# TODO this doesn't seem to have any effect, because we do not catch the exception anywhere?
+# https://docs.celeryq.dev/en/latest/userguide/configuration.html#task-soft-time-limit
 CELERY_TASK_SOFT_TIME_LIMIT = config(
```
stevenbal, line 89: @annashamray do you know more about this? In the Celery docs, it is mentioned that after this soft time limit a `SoftTimeLimitExceeded` is raised, and @Viicos mentioned that this time limit is needed for `notifications_api_common`, but I don't see that exception being caught in that library

docs/installation/config.rst
annashamray: I don't think so

src/objects/conf/base.py
annashamray: I've found the related PR (https://github.com/maykinmedia/objects-api/pull/362) but it's not much of help.
My guest that it is a copy-paste from Open Forms and I also haven't found any catching of `SoftTimeLimitExceeded`

stevenbal: In that case I'll remove this envvar entirely

src/objects/conf/base.py
```diff
@@ -107,22 +100,24 @@
 
 # setup_configuration command
 # sites config
-SITES_CONFIG_ENABLE = config("SITES_CONFIG_ENABLE", default=True)
-OBJECTS_DOMAIN = config("OBJECTS_DOMAIN", "")
-OBJECTS_ORGANIZATION = config("OBJECTS_ORGANIZATION", "")
+SITES_CONFIG_ENABLE = config("SITES_CONFIG_ENABLE", default=True, add_to_docs=False)
```
annashamray, line 103: Hmm I feel like a missing opportunity that we won't update `django-setup-configuration` config documentation with `generate_envvar_docs` command. 

Should we run a command to update it also inside `bin/generate_envvar_docs.sh`  ? 
Oh maybe in the future we can generate a separate page for them with it? 
Do you think this makes sense or I overcomplicate it?

stevenbal: I did not include the setup_configuration envvars yet, because there already was detailed documentation for them and they're not shared between projects. My goal here was mainly to remove the need to duplicate the same help texts for the shared variables.

I do think it would be good to include the setup_configuration envvars as well at some point, though I do think it's best to keep it as a separate page (like it is now) to differentiate between them.

I'm not yet sure what the easiest way to do that, I guess we could indeed run the command with a different `--file` and exclude the other groups. We'd probably also have to specify another template for it though, to add the intro text etc. Maybe it should be a separate issue?

annashamray: Yeah, and we can investigate in the separate issue that maybe Paul's doc generation is enough and covers all our needs?

annashamray (approved):
Please create an issue to generate document page for django_setup_configuration settings

Coperh (approved):

src/objects/conf/base.py
stevenbal: created https://github.com/maykinmedia/open-api-framework/issues/64 for it

PR #427, opened 2024-07-26
:stethoscope: add celery healthcheck

### LINKED ISSUE #424: Celery support health checks
### Thema / Theme

Objecten API

Same as: https://github.com/open-formulieren/open-forms/commit/5859733086a235cc1d0ce73d8f8868442b32a0cf#diff-10e7399873dcb65316dc6e44058e7b1fbff37e1f7c5b86731de852a274def9f6

_No response_

_No response_

sjoerdie:
Fixes #424 

**Changes**

- Add celery health check (Same as open_inwoner and openforms)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -76,12 +76,6 @@ services:
       start_period: 10s    
     depends_on:
       web:
-        # This health check condition is needed because Celery Beat will 
-        # try to convert the CELERY_BEAT_SCHEDULE into database entries. For 
-        # this, migrations need to be finished. If Celery tasks were still 
-        # pending, the database also needs to be ready for Celery itself. We 
-        # therefore have the health check here, and make Celery beat and 
-        # monitor containers depend on the celery container.
         condition: service_healthy
     volumes: *web_volumes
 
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -59,9 +68,21 @@ services:
...
+      test: ["CMD", "python", "/app/bin/check_celery_worker_liveness.py"]
+      interval: 30s
+      timeout: 5s
+      retries: 3
+      start_period: 10s    
     depends_on:
-      - db
-      - redis
+      web:
+        # This health check condition is needed because Celery Beat will
```
annashamray, line 79: We don't have celery beat for Objects API, so I think this comment is not needed

docker-compose.yml
```diff
@@ -36,6 +36,15 @@ services:
       - DEMO_TOKEN=demo-random-string
       - DEMO_PERSON=Demo
       - DEMO_EMAIL=demo@demo.local
+    healthcheck:
```
annashamray, line 39: Perhaps this check can be moved to web-init service, because "web" service depends on it

docker-compose.yml
sjoerdie: Oke, I dont even think the comment is correct.

docker-compose.yml
sjoerdie: Better leave it here, the init container will go down anyway when its done, also celery depends on the "web" service health check: 
```
    depends_on:
      web:
        condition: service_healthy
```

annashamray (approved):

Issue #328, opened 2023-05-08
As a developer I would like to get a specific version of an object


ferdi-ritense:
At this moment it is only possible to return the complete history of an object using `/api/v2/objects/{uuid}/history`. For objects with a large number of versions this is quite a heavy call. Often we already know the specific version we want and don't need all the other ones.

It would be really helpful to be able to request a specific version of an object, for example using `/api/v2/objects/{uuid}/history/{version}`

sdegroot:
Probably it should be `GET /api/v2/objects/{uuid}/{version}`

joeribekker:
Planned to be released in end of August.

PR #429, opened 2024-08-02
[#328] get specific object version

SonnyBA:
Fixes [#328](https://github.com/maykinmedia/objects-api/issues/328)

**Changes**

Allows retrieving objects by their `uuid` and `index`


src/objects/api/v2/openapi.yaml
```diff
@@ -457,6 +457,53 @@ paths:
...
+          enum:
+          - EPSG:4326
+        description: 'The desired ''Coordinate Reference System'' (CRS) of the response
+          data. According to the GeoJSON spec, WGS84 is the default (EPSG: 4326 is
+          the same as WGS84).'
+      - in: path
+        name: index
+        schema:
+          type: string
+          pattern: ^\d+$
```
stevenbal, line 478: Is this correct? I think this should be a positive integer instead of a string

stevenbal (changes requested):

annashamray (approved):
Same question as Steven's. Otherwise looks good

src/objects/api/v2/openapi.yaml
SonnyBA: See https://github.com/maykinmedia/objects-api/pull/429/commits/a9a808aae65bfa80290bed133fddb83df224d002

stevenbal (approved):

PR #428, opened 2024-08-01
:arrow_up: [#50] updated dependecies and fixed broken oidc tests

bart-maykin:
Fixes [#50](https://github.com/maykinmedia/open-api-framework/issues/50)

**Changes**

updated libraries:
- billiard `3.6.4.0` to `4.2.0`
- cbor2 `5.6.1` to `5.6.2`
- celery `5.2.7` to `5.4.0`
- certifi `2020.12.5` to `2023.7.22`
- django `4.2.11` to `4.2.15`
- django-axes `6.3.0` to `6.5.1`
- django-cors-headers `4.3.1` to `4.4.0`
- django-filters `23.5` to `24.2`
- django-jsonform `2.21.4` to `2.22.0`
- djangorestframework `3.14.0` to `3.15.2`
- drf-spectacular `0.27.0` to `0.27.2`
- elastic-apm `6.20.0` to `6.23.0`
- idna `2.10` to `3.7`
- jinja2 `3.1.3` to `3.1.4`
- maykin-2fa `1.0.0` to `1.0.1`
- mozilla-django-oidc-db `0.14.1` to `0.19.0`
- open-api-framework `0.5.0` to `0.6.1`
- python-dateutil `2.8.1` to `2.9.0.post0`
- requests `2.31.0` to `2.32.3`
- sentry-sdk `1.39.2` to `2.12.0`
- sqlparse `0.4.2` to `0.5.0`
- tornado `6.4` to `6.4.1`
- urllib3 `2.2.1` to `2.2.2`

updated oidc tests to work for new library version

PR #434, opened 2024-08-14
:arrow_up: upgraded zgw-consumers and removed zds-client from project

bart-maykin:
Fixes #428 

**Changes**

Updated zgw-consumers to 0.29.0 to update more python security issues
removed gemma-zds-client functionality thoughout the project

updated insecure libraries:
- certifi
- cryptography

### DIFF
--- a/src/objects/tests/v2/test_validation.py
+++ b/src/objects/tests/v2/test_validation.py
@@ -1,5 +1,6 @@
 import uuid
 
+import requests
 import requests_mock
 from rest_framework import status
 from rest_framework.test import APITestCase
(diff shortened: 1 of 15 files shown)


src/objects/tests/v2/test_validation.py
```diff
@@ -101,7 +101,7 @@ def test_create_object_no_version(self, m):
 
         data = response.json()
         self.assertEqual(
-            data["non_field_errors"], ["Object type version can not be retrieved: None"]
+            data["non_field_errors"], ["Object type doesn't have retrievable data"]
```
stevenbal, line 104: Same comment as in the v1 validation tests

stevenbal (changes requested):
Small comment about tests, looks good otherwise

stevenbal (approved):

PR #435, opened 2024-08-19
:bug: [#329] paginate object_history endpoint

### LINKED ISSUE #329: object_history endpoint is not actually paginated
The documentation shows the option to add page and pageSize parameters to the `/api/v2/objects/{uuid}/history` call. The example shows a paginated response.

However at the moment it seems that the response is always a complete unpaginated list of all object versions.

annashamray:
Fixes #329


Coperh (approved):
Looks good

PR #437, opened 2024-08-20
Documented required CACHE_DEFAULT environment parameter

Michel-Verhoeven:
Since objects-api:2.3.1 and objecttypes-api:2.2.0 a Redis instance is required for caching the backend. Since then the CACHE_DEFAULT env parameter is required which was not yet documented.


annashamray:
`CACHE_DEFAULT` is documented in the latest version - https://objects-and-objecttypes-api.readthedocs.io/en/latest/installation/config.html

PR #438, opened 2024-08-23
:arrow_up: bump open-api-framework to 0.8.0

annashamray:
After the PR is merged it would be added to patch release


docker-compose.yml
```diff
@@ -30,7 +30,8 @@ services:
       # setup_configuration env vars
       - OBJECTS_DOMAIN=web:8000
       - OBJECTS_ORGANIZATION=Objects
-      - OBJECTTYPES_API_ROOT=https://objecttypes.example.com/api/v2/
+      - OBJECTS_OBJECTTYPES_CONFIG_ENABLE=false
+      # - OBJECTTYPES_API_ROOT=https://objecttypes.example.com/api/v2/
```
stevenbal, line 34: Is it intentional to disable this?

stevenbal (commented):
@annashamray will you also add the warning about HSTS_SECONDS to the changelog? https://github.com/maykinmedia/open-api-framework/pull/63#discussion_r1728658426

docker-compose.yml
annashamray: Yes it's a temporary fix for quick-start action
`OBJECTTYPES_API_ROOT` setting was incorrect (it's fixed in this PR), but quick-start doesn't use the current build, but only the published one. 
So to make it work for now `OBJECTTYPES_API_ROOT` should be empty

annashamray:
@stevenbal I planed to add the warning when preparing the release. Do you want me to add it now?

stevenbal:
@annashamray no that's fine, was just to double check so we don't forget about it :+1:

stevenbal (approved):

PR #440, opened 2024-08-27
:sparkles:[#414] add in operator to API

### LINKED ISSUE #414: Support the `__in` operator in `data_attrs` to match one element
Poor man's OR-solution for https://github.com/maykinmedia/objects-api/issues/411 but covers @JanBrek use case.

Coperh:
Fixes #414

### DIFF
--- a/src/objects/api/constants.py
+++ b/src/objects/api/constants.py
@@ -9,3 +9,4 @@ class Operators(models.TextChoices):
     lt = "lt", _("lower than")
     lte = "lte", _("lower than or equal to")
     icontains = "icontains", _("case-insensitive partial match")
+    in_list = "in", _("contains")
--- a/src/objects/api/constants.py
+++ b/src/objects/api/constants.py
@@ -9,4 +9,4 @@ class Operators(models.TextChoices):
     lt = "lt", _("lower than")
     lte = "lte", _("lower than or equal to")
     icontains = "icontains", _("case-insensitive partial match")
-    in_list = "in", _("contains")
+    in_list = "in", _("in a list of values seperated by `|`")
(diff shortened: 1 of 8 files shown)


src/objects/tests/v1/test_filters.py
```diff
@@ -328,6 +328,31 @@ def test_filter_exclude_old_records(self):
...
+            data={"name": "demo1"}, object__object_type=self.object_type
+        )
+        record2 = ObjectRecordFactory.create(
+            data={"name": "demo2"}, object__object_type=self.object_type
+        )
+        ObjectRecordFactory.create(
+            data={"name": "demo3"}, object__object_type=self.object_type
+        )
+
+        response = self.client.get(self.url, {"data_attrs": "name__in__demo1|demo2"})
```
Coperh, line 342: Don't really like this format, but I cannot think of a better one. The comma is already used for multiple attributes.

src/objects/api/constants.py
```diff
@@ -9,3 +9,4 @@ class Operators(models.TextChoices):
     lt = "lt", _("lower than")
     lte = "lte", _("lower than or equal to")
     icontains = "icontains", _("case-insensitive partial match")
+    in_list = "in", _("in a list of values seperated by `|`")
```
stevenbal, line 12:
```suggestion
    in_list = "in", _("in a list of values separated by `|`")
```

src/objects/tests/v1/test_filters.py
stevenbal: Hmm yeah, not sure what alternatives there are

stevenbal (changes requested):
Looks good overall

stevenbal (approved):

PR #444, opened 2024-09-06
don't show previous records available on some date

### LINKED ISSUE #324: Search still returns previous versions of objects for a while after they have been updated

annashamray:
Fixes #324

### DIFF
--- a/src/objects/tests/v2/test_object_api.py
+++ b/src/objects/tests/v2/test_object_api.py
@@ -20,7 +20,7 @@ from objects.utils.test import TokenAuthMixin
 
 from ..constants import GEO_WRITE_KWARGS
 from ..utils import mock_objecttype, mock_objecttype_version, mock_service_oas_get
-from .utils import reverse
+from .utils import reverse, reverse_lazy
 
 OBJECT_TYPES_API = "https://example.com/objecttypes/v1/"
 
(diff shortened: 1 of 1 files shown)


src/objects/api/v2/filters.py
```diff
@@ -124,7 +124,9 @@ def filter_data_icontains(self, queryset, name, value: str):
         return queryset.extra(where=[where_str], params=[value])
 
     def filter_date(self, queryset, name, value: date_):
-        return queryset.filter_for_date(value)
+        """actual filtering is performed in the viewset"""
```
bart-maykin, line 127: What does this do? From what I can see all this does is return the queryset. Can't we just remove this?

src/objects/api/v2/filters.py
```diff
@@ -124,7 +124,9 @@ def filter_data_icontains(self, queryset, name, value: str):
...
+        return queryset
 
     def filter_registration_date(self, queryset, name, value: date_):
-        return queryset.filter_for_registration_date(value)
+        """actual filtering is performed in the viewset"""
```
bart-maykin, line 131: Same as question as filter_date

Coperh (approved):
Its confusing, but I think its fine.

src/objects/api/v2/filters.py
annashamray: It's added to the filterset to generate the OAS with correct query params
But the actual filtering is moved to the viewset now, because it should be done before filtering on other query params

src/objects/api/v2/filters.py
annashamray: Same thing - It's added to the filterset just to generate the OAS with correct query params

src/objects/tests/v2/test_object_api.py
```diff
@@ -430,3 +430,134 @@ def test_update_object_correctionFor(self, m):
...
+            self.assertEqual(object_data["record"]["data"], {"name": "old"})
+
+        with self.subTest("without filter on old name"):
+            response = self.client.get(self.url, {"registrationDate": "2024-08-31"})
+
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+            self.assertEqual(response.json()["count"], 1)
+            object_data = response.json()["results"][0]
+            self.assertEqual(object_data["uuid"], str(self.object.uuid))
+            self.assertEqual(object_data["record"]["data"], {"name": "new"})
```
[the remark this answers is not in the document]
annashamray: I tried to describe it in the dosctring 
The idea is if two object records are available on a particular date, only the latest one should be always shown or no record should be shown (depending on query params)
So the first record should never appeared in the response for this particular date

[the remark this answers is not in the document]
annashamray: On the second thought I've noticed that the test name at line 487 is not correct, I'll fix it

PR #452, opened 2024-09-19
:bug: [open-zaak/open-zaak#1780] Fix setting NOTIFICATIONS_DISABLED via envvar

stevenbal:
Fixes open-zaak/open-zaak#1780 partially

**Changes**

* Fix setting NOTIFICATIONS_DISABLED via envvar


annashamray (approved):
I like the change

PR #454, opened 2024-09-24
:bug:[#449] support objecttypes API with pagination

### LINKED ISSUE #449: Objects API: setting permissions causes a 500 error

latest?

https://sentry.maykinmedia.nl/organizations/maykin-media/issues/371363/?project=428&query=is%3Aunresolved

Version seems unknown? Install is from a few days ago

image

_No response_

_No response_

Coperh:
Fixes #449

Supports paginated response from OT V2

### DIFF
--- a/src/objects/token/admin.py
+++ b/src/objects/token/admin.py
@@ -52,6 +52,9 @@ class PermissionAdmin(admin.ModelAdmin):
             except requests.JSONDecodeError:
                 continue
 
+            if "results" in response_data:
+                response_data = response_data["results"]
+
             # use only first level of properties
             data_fields[object_type.id] = {
                 version["version"]: {
--- a/src/objects/token/admin.py
+++ b/src/objects/token/admin.py
@@ -52,6 +52,7 @@ class PermissionAdmin(admin.ModelAdmin):
             except requests.JSONDecodeError:
                 continue
 
+            # TODO: remove check once API V1 is removed
             if "results" in response_data:
                 response_data = response_data["results"]
 
--- a/src/objects/token/tests/test_admin.py
+++ b/src/objects/token/tests/test_admin.py
@@ -1,0 +1,79 @@
+from django.test import TestCase, tag
+from django.urls import reverse
+
+from maykin_2fa.test import disable_admin_mfa
+from vcr.unittest import VCRMixin
+from zgw_consumers.constants import AuthTypes
+from zgw_consumers.test.factories import ServiceFactory
+
+from objects.accounts.tests.factories import UserFactory
+from objects.core.tests.factories import ObjectTypeFactory
+
+
+@disable_admin_mfa()
+class PermissionAdminTests(VCRMixin, TestCase):
+
+    object_types_api = "http://127.0.0.1:8008/api/{version}/"
+
+    def setUp(self):
+        super().setUp()
+
+        self.user = UserFactory(superuser=True)
+        self.client.force_login(self.user)
+        self.url = reverse("admin:token_permission_add")
+
+    def _get_vcr(self, **kwargs):
+        vcr = super()._get_vcr(**kwargs)
+        vcr.filter_headers = ["authorization"]
+        return vcr
+
+    @tag("#449")
+    def test_with_object_types_api_v1(self):
+        """
+        Regression test for #449.
+        Test if Permission admin can still handle objecttypes API V1
+        """
+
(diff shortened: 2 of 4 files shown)


src/objects/token/tests/test_admin.py
```diff
@@ -0,0 +1,76 @@
...
+    def test_with_object_types_api_v2(self):
+        """
+        Regression test for #449.
+        Test if Permission admin can handle objecttypes API V2 which added pagination
+        """
+        v2_service = ServiceFactory(
+            api_root=self.object_types_api.format(version="v2"),
+            auth_type=AuthTypes.api_key,
+            header_key="Authorization",
+            header_value="Token 5cebbb33ffa725b6ed5e9e98300061218ba98d71",
```
Coperh, line 64: Should a docker container be made for objecttypes for testing here?

Coperh:
Idk why this is failing, I'll look into it later

src/objects/token/tests/test_admin.py
annashamray: I think so. We should be able to change and modify these tests. 
If you decide to use vcr let's make sure we can reuse it:
* please add docker-compose with Objecttypes
* add fixture for Obejcttypes if you use it
* add README which describes how to use it if you want to update cassettes

annashamray (commented):
I like the idea to using vcr for Objecttypes API, but I think it needs a little more love

src/objects/token/tests/test_admin.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+        ObjectTypeFactory(
+            service=v1_service, uuid="71a2452a-66c3-4030-b5ec-a06035102e9e"
+        )
+
+        response = self.client.get(self.url)
+
+        print(response)
```
stevenbal, line 49: Can be removed

src/objects/token/tests/test_admin.py
```diff
@@ -0,0 +1,79 @@
...
+        self.assertEqual(response.status_code, 200)
+
+        self.assertEqual(
+            self.cassette.requests[1].uri,
+            "http://127.0.0.1:8008/api/v1/objecttypes/71a2452a-66c3-4030-b5ec-a06035102e9e/versions",
+        )
```
stevenbal, line 55: Not sure if checking this is needed here, maybe you could add an assertion instead to check if the proper objecttypes are passed via response.context to populate the objecttype dropdown

stevenbal (changes requested):

src/objects/token/tests/test_admin.py
Coperh: I've split this into a separate issue #457

src/objects/token/admin.py
```diff
@@ -52,6 +52,10 @@ def get_data_field_choices(self):
             except requests.JSONDecodeError:
                 continue
 
+            # TODO: remove check once API V1 is removed
```
Coperh, line 55: This check will be pointless once Objectstypes  API  V1 is discontinued

src/objects/token/tests/test_admin.py
```diff
@@ -0,0 +1,84 @@
...
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+
+        form = response.context["adminform"].form
+        choices = list(form.fields["object_type"].choices)
+
+        self.assertEqual(
+            choices[1][0].value,
+            object_type.id,
+        )
```
Coperh, line 84: I feel like there was an easier way to check this with normal Django but maybe I'm thinking of webtest.

stevenbal: hmm yeah I think that's probably webtest

src/objects/token/tests/test_admin.py
```diff
@@ -0,0 +1,84 @@
...
+
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+
+        form = response.context["adminform"].form
+        choices = list(form.fields["object_type"].choices)
+
+        self.assertEqual(
+            choices[1][0].value,
+            object_type.id,
```
stevenbal, line 83: bit of a nitpick, but is it possible to check the label here as well? Currently it will probably just check if the `value` is `1`, which doesn't really say a lot :grimacing:

stevenbal (changes requested):

stevenbal (approved):

PR #460, opened 2024-10-01
:construction_worker: check OAF version on PRs

Coperh:
Fixes maykinmedia/open-api-framework#44

### DIFF
--- a/.github/workflows/oaf-check.yml
+++ b/.github/workflows/oaf-check.yml
@@ -20,8 +20,10 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.11'
+          cache: 'pip'
+          cache-dependency-path: 'requirements/dev.txt'
       - name: Install dependencies
-        run: pip install -U pip-tools
+        run: pip install $(grep "pip-tools==" requirements/dev.txt)
       - name: Run compile dependencies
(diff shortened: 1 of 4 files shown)


Coperh:
Needs #458 to be merged

Coperh:
Should I pin pip-tools or should I add it to the check OAF version action?

SonnyBA:
Why should it be pinned? It is currently added through `dev.in` which is correct right?

Coperh:
About this comment: https://github.com/maykinmedia/open-klant/pull/255#discussion_r1782710608

IF there is a stable branch, you might want to keep OAF at a certain version, so you might only want to keep master up to date.

.github/workflows/oaf-check.yml
```diff
@@ -14,8 +20,10 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.11'
+          cache: 'pip'
+          cache-dependency-path: 'requirements/dev.txt'
       - name: Install dependencies
-        run: pip install -U pip-tools
+        run: pip install -r requirements/dev.txt # use the same pip-tools as dev
```
stevenbal, line 26: We could do something like this to avoid installing unnecessary dependencies, `pip` isn't the fastest

```
pip install $(grep "pip-tools==" requirements/dev.txt)
```

stevenbal (approved):

PR #475, opened 2024-11-14
[#474] Fix version number

### LINKED ISSUE #474: Version number doesn't seem to be shown

2.4.3

**Objects API (2.4.3) isn't showing the version number**

Bug related: https://github.com/maykinmedia/open-api-framework/issues/82

_No response_

_No response_

danielmursa-dev:
Fixes #474 
Fixes https://github.com/maykinmedia/open-api-framework/issues/82

The version number is already correctly set up in [open-api-framework](https://github.com/maykinmedia/open-api-framework), but in [objects-api](https://github.com/maykinmedia/objects-api) PROD was not configured, so the change is to set the value of the environment variables ‘RELEASE’ with the value from api.py

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,6 +79,10 @@ ARG RELEASE
 ENV GIT_SHA=${COMMIT_HASH}
 ENV RELEASE=${RELEASE}
 
+RUN echo "***********************************************"
+RUN echo "GIT_SHA=${GIT_SHA}" && echo "RELEASE=${RELEASE}"
+RUN echo "***********************************************"
+
 ENV DJANGO_SETTINGS_MODULE=objects.conf.docker
 
 ARG SECRET_KEY=dummy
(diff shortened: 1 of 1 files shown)


src/objects/conf/production.py
```diff
@@ -4,6 +4,7 @@
 
 from .base import *  # noqa isort:skip
 
+os.environ.setdefault("RELEASE", API_VERSION)
```
stevenbal, line 7: `API_VERSION` is the version of the API endpoint itself and we want to show the version of the application in the admin. This should be the `RELEASE` setting defined in `open-api-framework` https://github.com/maykinmedia/open-api-framework/blob/main/open_api_framework/conf/base.py#L640. I'm not sure why it isn't picked up here though, maybe it's not being set in the Docker image?

stevenbal (changes requested):

PR #488, opened 2024-11-29
support commas in data_attrs query param

### LINKED ISSUE #472: zoeken op waarden met een komma erin, geeft een foutmelding

Objects API (2.3.1 (v2))

Als je in de Objectenregistratie een object probeert op te halen op een tekststring waarin een komma zit, dan is de repsonse: 
400 Bad request

response body: 
```
[
    "not enough values to unpack (expected 3, got 1)"
]
```

Zie ook screenshot uit Postman met een aantal calls, na elkaar gedaan. Allemaal geven ze dezelfde response. We hebben deze calls geprobeerd

https://objecten.dev.kiss-demo.nl/api/v2/objects?data_attrs=naam__icontains__Advies%2C+support+en+kennis
https://objecten.dev.kiss-demo.nl/api/v2/objects?data_attrs=naam__icontains__Advies, support en kennis
https://objecten.dev.kiss-demo.nl/api/v2/objects?data_attrs=naam__icontains__Advies%2C+support+en+kennis+%28ASK%29
https://objecten.dev.kiss-demo.nl/api/v2/objects?data_attrs=naam__icontains__Advies, support en kennis (ASK)

NB; deze gaan wel goed: 
https://objecten.dev.kiss-demo.nl/api/v2/objects?data_attrs=naam__icontains__Sociaal+Wijkteam
https://objecten.dev.kiss-demo.nl/api/v2/objects?data_attrs=naam__icontains__Advies+support+en+kennis+%28ASK%29

image-20241106-203014

(Versienr van de API gekopieerd vanaf https://objecten.dev.kiss-demo.nl/api/v2/schema/)

Wel kunnen zoeken op strings waar een komma in staat

annashamray:
Fixes #472 

We use comma as a separator between `data_attrs` values
If we want to support commas inside values we can consider several options:

1. Remove several value groups, separated by commas, so comma won't be a reserved character here
   Before: `?data_attrs=naam__icontains__anna,naam__icontains__Advies` 
   After:`?data_attrs=naam__icontains__anna&data_attrs=naam__icontains__Advies, support`

   This is my personal favorite option, but it will break current API
   
2. Change separator from comma to other character, for example to `|`. 
   Before: `?data_attrs=naam__icontains__anna,naam__icontains__Advies` 
   After: `?data_attrs=naam__icontains__anna|naam__icontains__Advies, support` 
   
   Also a nice option, but also breaks current API
   
3. Escape comma in the value part or use other char instead of comma in the value
   Here comma in the value is replaced with `\,`
   Before: `?data_attrs=naam__icontains__anna,naam__icontains__Advies` 
   After: `?data_attrs=naam__icontains__anna,naam__icontains__Advies\, support` 
   
   This options is implemented in the PR, since it doesn't break the API
   
4. Use some regex magic to define when comma is used as separator and when it's used as a part of value. 
   Let's not do it, since it's very tricky and unclear what happens

What do you think is the best way to implement it? 
Should we introduce the breaking change? The next big release already contains a breaking change, since we remove v1 endpoints 

@joeribekker @stevenbal @SonnyBA @Coperh let's discuss it

### DIFF
--- a/src/objects/api/v2/openapi.yaml
+++ b/src/objects/api/v2/openapi.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 info:
   title: Objects API
-  version: 2.4.3 (v2)
+  version: 2.4.3
   description: |
     An API to manage Objects.
 
@@ -89,12 +89,12 @@ paths:
           data. According to the GeoJSON spec, WGS84 is the default (EPSG: 4326 is
           the same as WGS84).'
       - in: query
-        name: data_attrs
+        name: data_attr
         schema:
           type: string
         description: |
           Only include objects that have attributes with certain values.
-          Data filtering expressions are comma-separated and are structured as follows:
+
           A valid parameter value has the form `key__operator__value`.
           `key` is the attribute name, `operator` is the comparison operator to be used and `value` is the attribute value.
           Note: Values can be string, numeric, or dates (ISO format; YYYY-MM-DD).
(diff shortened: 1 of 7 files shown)


src/objects/api/v2/openapi.yaml
```diff
@@ -88,13 +88,51 @@ paths:
...
+          should be used. If `height` is nested inside `dimensions` attribute, query should look like
+          `data_attr=dimensions__height__exact__100`
+
+          This filter is very similar to the old `data_attrs` filter, but it has two differences:
+
+          * `value` may contain commas
+          * only one filtering expression is allowed
+
+          If you want to use several filtering expressions, just use this `data_attr` several times in the query string.
+          Example: `data_attr=height__exact__100&data_attr=naam__icontains__boom`
```
stevenbal, line 126: Nitpicky, but ideally `explode: true` would be specified here to indicate that it should be used like `?data_attr=...&data_attr=...` (see https://swagger.io/docs/specification/v3_0/serialization/#query-parameters)

For example: https://github.com/open-formulieren/open-forms/blob/3a702080d35c0670588cc69b8e49445e6224989c/src/openforms/appointments/api/views.py#L64

Although looking at it, apparently `explode: true` is considered the default by OAS3, so I'm not sure if our schema is correct in that case :thinking:. Might be something to investigate in another issue

src/objects/api/validators.py
```diff
@@ -68,32 +68,54 @@ def __call__(self, new_value, serializer_field):
...
+    code = "invalid-data-attr-query"
+
+    for value_part in value:
+        # check that comma can be only in the value part
+        if "," in value_part.rsplit("__", 1)[0]:
             message = _(
-                "Operator `%(operator)s` supports only dates and/or numeric values"
-            ) % {"operator": operator}
+                "Filter expression '%(value_part)s' doesn't have the shape 'key__operator__value'"
+            ) % {"value_part": value_part}
```
stevenbal, line 114: Since this error specifically occurs when there are commas outside the value part, maybe it would be good to explicitly mention that that isn't allowed in the error, something like: `Filter expression '%(value_part)s' must have the shape 'key__operator__value', commas can only be present in the 'value'`

src/objects/tests/v2/test_filters.py
```diff
@@ -428,6 +432,393 @@ def test_filter_in_string(self):
...
+        self.assertEqual(len(data), 1)
+        self.assertEqual(
+            data[0]["url"],
+            f"http://testserver{reverse('object-detail', args=[record.object.uuid])}",
+        )
+
+    def test_filter_exact_date(self):
+        record = ObjectRecordFactory.create(
+            data={"date": "2000-11-01"}, object__object_type=self.object_type
+        )
```
stevenbal, line 494: minor, but since you also add another object in the other tests that doesn't match the filter value, it might be good to do it here as well (to verify that the filter doesn't return all objects)

stevenbal (changes requested):
Minor comments, looks good overall :+1:

stevenbal (approved):

PR #494, opened 2024-12-16
[#485] tokenauth setup configuration

### LINKED ISSUE #485: Add support for Token configuration via django-setup-configuration

danielmursa-dev:
Fixes #485
Without permissions


joeribekker:
We can split off permissions so this can already be merged, which results in a working yaml file (although every token has superuser permissions for now).

src/objects/conf/base.py
```diff
@@ -84,8 +84,9 @@
 # Django setup configuration
 #
 SETUP_CONFIGURATION_STEPS = (
+    "django_setup_configuration.contrib.sites.steps.SitesConfigurationStep",
```
stevenbal, line 87: Since this step was also added, could you also add data for this to `docker/setup_configuration/data.yml`? 

```yaml
sites_config_enable: true
sites_config:
  items:
  - domain: example.com
    name: Example site
```

stevenbal (approved):
Just a comment about an example in `data.yml`, looks good otherwise

PR #508, opened 2024-12-31
Feature/464 permission admin view slow

### LINKED ISSUE #464: Objecten API: Permission admin view is slow

2.4.4

Taiga DH 596

Adding a new permission in the Objects API for DH currently takes ~12 seconds, this because of the retrieval of all the objecttypes on every pageload.

For further details and an APM trace please see the Taiga issue

_No response_

_No response_

danielmursa-dev:
Fixes #464 

**Changes**

From ticket comments to solve the slow page loading, the option was chosen not to load all the fields of each `objecttype ` unnecessarily, because only one `objecttype` is to be used at a time, so it was decided to only load them when the `objecttype` is selected and the `use_fields` flag is also enabled.

The initial option I had thought of was to do this on the FE side so that I could make single requests for each `objecttype`, but this is more difficult because I would have to pass to the view context more elements such as `uuid`, `service data`, `token` eg. of the `object_type` to then form, for a correctly and secure request;
So I decided to add a simple control that in order to modify the fields of the `permission` must exist and the `use_fields` must also be enabled. 
This way, when the page is reloaded, the permission has all the data and can directly retrieve the information about the fields choices.

image

### DIFF
--- a/src/objects/js/components/admin/permissions/permission-form.js
+++ b/src/objects/js/components/admin/permissions/permission-form.js
@@ -9,8 +9,10 @@ const PermissionForm = ({objectFields, dataFieldChoices, tokenChoices, objecttyp
     const [mode, setMode]  = useState(values["mode"]);
     const [useFields, setUseFields] = useState(values["use_fields"]);
     const [objectType, setObjectType] = useState(values["object_type"]);
-
-    const [fields, setFields] = useState( JSON.parse(values["fields"]) || {});
+    if (!values["fields"]) {
+        values["fields"] = "{}"
+      }
+    const [fields, setFields] = useState( JSON.parse(values["fields"]) || {})
 
     return (
      <fieldset className="module aligned">
@@ -63,7 +65,7 @@ const PermissionForm = ({objectFields, dataFieldChoices, tokenChoices, objecttyp
                 name="use_fields"
                 id="id_use_fields"
                 label="Use field-based authorization"
-                disabled={mode === "read_and_write" || Object.keys(dataFieldChoices).length === 0}
+                disabled={!mode || mode === "read_and_write"}
                 value={useFields}
                 onChange={(value) => {setUseFields(value)}}
             />
--- a/src/objects/js/components/admin/permissions/auth-fields.js
+++ b/src/objects/js/components/admin/permissions/auth-fields.js
@@ -61,6 +61,14 @@ const authFields = (object_fields, dataFields, fields, setFields) => {
 };
 
 const versionAuthFields = (objectType, objectFields, dataFieldChoices, fields, setFields) => {
+    if (!(objectType in dataFieldChoices)) {
+        return (
+            <div className="errornote">
+                <p>To be able to select fields, first you must save the instance and then you can edit the fields!</p>
+                <p>Click on button <strong>'Save and continue editing'</strong></p>
+            </div>
+        );
+    }
 
     const dataFields = dataFieldChoices[objectType];
     const objecttypeVersions = Object.entries(dataFieldChoices).reduce((acc, [k, v]) => {
--- a/src/objects/tests/admin/test_token_permissions.py
+++ b/src/objects/tests/admin/test_token_permissions.py
@@ -6,7 +6,12 @@ from maykin_2fa.test import disable_admin_mfa
 from requests.exceptions import ConnectionError
 
 from objects.accounts.tests.factories import UserFactory
-from objects.token.tests.factories import ObjectTypeFactory, TokenAuthFactory
+from objects.token.constants import PermissionModes
+from objects.token.tests.factories import (
+    ObjectTypeFactory,
+    PermissionFactory,
+    TokenAuthFactory,
+)
 
 from ..utils import mock_objecttype, mock_objecttype_version, mock_service_oas_get
 
@@ -37,18 +42,7 @@ class AddPermissionTests(WebTest):
         response = self.app.get(self.url)
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(
-            response.context["data_field_choices"],
-            {
-                object_type.id: {
-                    1: {},
-                    2: {
-                        "diameter": "record__data__diameter",
-                        "plantDate": "record__data__plantDate",
-                    },
-                }
-            },
-        )
+        self.assertEqual(response.context["data_field_choices"], {})
 
     def test_get_permission_with_unavailable_objecttypes(self, m):
         """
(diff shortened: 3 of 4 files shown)


stevenbal (changes requested):
I was thinking of the following to make it possible to fetch the fields via the frontend (such that the form doesn't have to be saved to fetch them):
- make an endpoint on the admin (that requires admin authentication) that performs the request for a specific object_type_id to retrieve the `data_fields` (e.g. `/admin/objecttypes/<id>`). That way we don't have to expose the token to the frontend and we only need to pass the `id` to the frontend
- in react, call this new endpoint for the currently selected objecttype and display the result

@annashamray what do you think about this?

src/objects/core/admin.py
```diff
@@ -13,6 +18,32 @@ class ObjectTypeAdmin(admin.ModelAdmin):
...
+        urls = super().get_urls()
+        my_urls = [
+            path(
+                "<int:objecttype_id>/versions/",
+                self.admin_site.admin_view(self.versions_view),
+            )
+        ]
+        return my_urls + urls
+
+    def versions_view(self, request, objecttype_id):
```
stevenbal, line 31: Could you add two tests to verify that non authenticated and non staff users cannot access this endpoint?

src/objects/core/admin.py
```diff
@@ -13,6 +18,32 @@ class ObjectTypeAdmin(admin.ModelAdmin):
     )
     readonly_fields = ("_name",)
 
+    def get_urls(self):
+        urls = super().get_urls()
+        my_urls = [
+            path(
+                "<int:objecttype_id>/versions/",
```
stevenbal, line 25: Since it's for internal use only, let's prefix it with an underscore
```suggestion
                "<int:objecttype_id>/_versions/",
```

src/objects/core/admin.py
```diff
@@ -13,6 +18,32 @@ class ObjectTypeAdmin(admin.ModelAdmin):
...
+            "count": 0,
+            "next": None,
+            "previous": None,
+            "results": [],
+        }
+        if objecttype := self.get_object(request, objecttype_id):
+            client = build_client(objecttype.service)
+            try:
+                response = client.get(objecttype.versions_url)
+                versions = response.json()
```
stevenbal, line 42: This doesn't handle pagination currently, I think it's best to do instantiate `versions = []` and then in the code below use the `pagination_helper` from `zgw-consumers` to return all the results: https://zgw-consumers.readthedocs.io/en/latest/recipes.html#handling-pagination-in-api-endpoints

src/objects/core/admin.py
```diff
@@ -13,6 +18,32 @@ class ObjectTypeAdmin(admin.ModelAdmin):
...
+            except (requests.RequestException, requests.JSONDecodeError):
+                pass
```
stevenbal, line 44: Let's at least emit a log here
```suggestion
                logger.exception("Something went wrong while fetching objecttype versions")
```

src/objects/js/components/admin/permissions/permission-form.js
```diff
@@ -63,7 +98,7 @@ const PermissionForm = ({objectFields, dataFieldChoices, tokenChoices, objecttyp
                 name="use_fields"
                 id="id_use_fields"
                 label="Use field-based authorization"
-                disabled={mode === "read_and_write" || Object.keys(dataFieldChoices).length === 0}
+                disabled={!mode || mode === "read_and_write"}
```
stevenbal, line 101: Let's also make it disabled if the objecttypes API is not available, currently you can check it despite that:

image

src/objects/js/components/admin/permissions/permission-form.js
```diff
@@ -35,6 +69,7 @@ const PermissionForm = ({objectFields, dataFieldChoices, tokenChoices, objecttyp
                 errors={errors["object_type"]}
                 onChange={(value) => {
                     setObjectType(value);
+                    fetchObjecttypeVersions(value);
```
stevenbal, line 72: Is this needed? Currently two requests are made when the objecttype is changed, I think the `useEffect` hook above is probably sufficient.

src/objects/js/components/admin/permissions/permission-form.js
```diff
@@ -1,17 +1,51 @@
...
-
+    if (!values["fields"]) {
+        values["fields"] = "{}"
+    }
+
+    const [fields, setFields] = useState( JSON.parse(values["fields"]) || {})
+    const [dataFieldChoices, setDataFieldChoices] = useState(dataFieldChoices);
+
+    const fetchObjecttypeVersions = (objecttype_id) => {
+        fetch(`/admin/core/objecttype/`+ objecttype_id +`/versions/`, {
```
stevenbal, line 20:
```suggestion
        fetch(`/admin/core/objecttype/${objecttype_id}/versions/`, {
```

stevenbal (changes requested):
Looks good overall, some minor questions/remarks

src/objects/core/admin.py
```diff
@@ -13,6 +23,30 @@ class ObjectTypeAdmin(admin.ModelAdmin):
...
+        versions = []
+        if objecttype := self.get_object(request, objecttype_id):
+            client = build_client(objecttype.service)
+            try:
+                response = client.get(objecttype.versions_url)
+                versions = list(pagination_helper(client, response.json()))
+            except (requests.RequestException, requests.JSONDecodeError):
+                logger.exception(
+                    "Something went wrong while fetching objecttype versions"
+                )
```
annashamray, line 47: Could you please add test what happens if the exception is raised here?

src/objects/tests/admin/test_core_views.py
```diff
@@ -0,0 +1,32 @@
...
+            response,
+            f"You are authenticated as {user.username}, but are not authorized to access this page",
+        )
+
+        user.is_staff = True
+        user.save()
+
+        url = reverse("admin:objecttype_versions", args=[1])
+        response = self.app.get(url, user=user)
+        self.assertEqual(response.json, [])
```
annashamray, line 32: It looks like this test can be separated to two tests:
* check permissions for non-staff use
* check happy flow

Could you add data in the response of the happy flow? So we know that the new endpoint works as expected

src/objects/js/components/admin/permissions/permission-form.js
```diff
@@ -1,17 +1,51 @@
...
     const [objectType, setObjectType] = useState(values["object_type"]);
-
-    const [fields, setFields] = useState( JSON.parse(values["fields"]) || {});
-
+    if (!values["fields"]) {
+        values["fields"] = "{}"
+    }
+
+    const [fields, setFields] = useState( JSON.parse(values["fields"]) || {} )
+    const [dataFieldChoices, setDataFieldChoices] = useState(dataFieldChoices);
```
annashamray, line 17:
```suggestion
    const [dataFieldChoices, setDataFieldChoices] = useState({});
```
Initial state should be different from the current state

annashamray (changes requested):
One thing I've noticed now, but it's been implemented like this since the beginning:

The state of selected fields is not tracked:
If you change the objecttype in the dropdown and then change it back, the fields previous will not be selected, even if they are saved in the DB
[Screencast from 28-01-25 16:32:04.webm](https://github.com/user-attachments/assets/3a6841e1-8eeb-48fb-aacf-8b0f1120c01d)

It's been like this probably forever, so could you please create an issue to fix it? 
No need to fix it in this PR, but some cleanup would be nice

src/objects/tests/admin/test_core_views.py
```diff
@@ -0,0 +1,69 @@
...
+        response = self.app.get(url, user=user)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.json), 0)
+
+    def test_endpoint_unreachable(self, m):
+        user = UserFactory.create(is_staff=True, is_superuser=True)
+        object_type = ObjectTypeFactory.create()
+        url = reverse("admin:objecttype_versions", args=[object_type.pk])
+
+        with self.assertRaises(requests_mock.exceptions.NoMockAddress):
```
stevenbal, line 53: This is an exception that won't happen outside of tests, you can specify `exc=...` to let `requests_mock` raise a specific exception for a URL:

```python
from requests.exceptions import HTTPError
m.get(f"{object_type.url}/versions", exc=HTTPError)
```

And I think in that case, the endpoint returns an empty list

src/objects/tests/admin/test_core_views.py
```diff
@@ -0,0 +1,69 @@
...
+        url = reverse("admin:objecttype_versions", args=[object_type.pk])
+        response = self.app.get(url, user=user)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.json), 1)
+
+        # object_type does not exist
+        url = reverse("admin:objecttype_versions", args=[100])
+        response = self.app.get(url, user=user)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.json), 0)
```
stevenbal, line 46: let's verify that it is an empty list (and not an object or something else with len == 0)

```suggestion
        self.assertEqual(response.json, [])
```

src/objects/tests/admin/test_core_views.py
```diff
@@ -0,0 +1,69 @@
...
+        user = UserFactory.create(is_staff=True, is_superuser=True)
+
+        # object_type exist
+        url = reverse("admin:objecttype_versions", args=[object_type.pk])
+        response = self.app.get(url, user=user)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.json), 1)
+
+        # object_type does not exist
+        url = reverse("admin:objecttype_versions", args=[100])
```
stevenbal, line 43: Dependent on the amount of objecttypes created in other tests (or using `--keepdb` locally) this PK might be valid, so let's make sure it is unused
```suggestion
        url = reverse("admin:objecttype_versions", args=[object_type.pk+1])
```

annashamray (approved):
Please create an issue for keeping track of selected fields after changing the objecttype

stevenbal (approved):
Some minor comments about the tests, looks good otherwise

PR #529, opened 2025-02-18
:heavy_plus_sign: Add bump-my-version to dev dependencies

stevenbal:
Fixes maykinmedia/open-api-framework#107 partially

**Changes**

* Add bump-my-version to dev dependencies

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -12,10 +12,8 @@ search = :Version: {current_version}
 replace = :Version: {new_version}
 
 [bumpversion:file:package.json]
-search = "version": "{current_version}",
-replace = "version": "{new_version}",
-
-[bumpversion:file:src/objects/conf/api.py]
+search = "version": "{current_version}"
+replace = "version": "{new_version}"
 
 [bumpversion:file:src/objects/__init__.py]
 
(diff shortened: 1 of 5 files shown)


.bumpversion.cfg
```diff
@@ -12,10 +12,8 @@ search = :Version: {current_version}
 replace = :Version: {new_version}
 
 [bumpversion:file:package.json]
-search = "version": "{current_version}",
-replace = "version": "{new_version}",
-
-[bumpversion:file:src/objects/conf/api.py]
```
stevenbal, line 18: @annashamray I think this was incorrectly part of bumpversion.cfg, it includes the API version which I think are separate from the version of the software, right?

annashamray: It's not an easy question actually. Maybe we need to discuss it with @joeribekker 
When we supported v1 and v2 of API at the same time we had Objects software version == Objects API v2 version

Now we dropped v1 of the API and released v3 of software and they started to deviate from each other. 
I personally don't like it. Objects API is mostly an Restful API with a little bit of admin interface, so it doesn't make a lot of sense to have separate versions. Moreover unlike OZ we don't have any standard to adhere to. 
I think having one version would be less confusing and more convenient for both developers and clients

But it's up to discussion

stevenbal: Let's discuss it when Joeri is back indeed

PR #569, opened 2025-04-15
:arrow_up: [maykinmedia/open-klant#341] Upgrade commonground-api-common

danielmursa-dev:
Upgrade `commonground-api-common`

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -57,7 +57,7 @@ click-plugins==1.1.1
     # via celery
 click-repl==0.3.0
     # via celery
-commonground-api-common==2.5.5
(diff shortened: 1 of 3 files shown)


stevenbal (commented):
@danielmursa-dev are the help texts for the filters here already explicitly defined? Since I don't see a diff for the schema

danielmursa-dev:
> @danielmursa-dev are the help texts for the filters in this API already explicitly defined? Since I don't see a diff for the schema

@stevenbal Yes, I think so, there is only one class `FilterSet`, where the fields are already defined explicitly and not by the model, so I think that is why the API spec is the same.

PR #571, opened 2025-04-22
:sparkles: [#550] Add cache to retrieve versions of objecttypes

### LINKED ISSUE #550: Avoid repeatedly retrieving the same objecttype when creating many new objects
### Thema / Theme

Objecten API

Taiga Utrecht 333

When creating many objects, each POST leads to a GET on the objecttypes API. 

"De error is afkomstig uit de controle of het objecttype/versie van het object bestaat, bij ieder nieuw (of update van bestaand) object zal de objecttype API worden aangeroepen."

"de situatie doet zich sporadisch voor, maar als het zich voordoet dan krijgen we zeer veel van dit soort foutmeldingen"

Caching for a short term of the objecttype would seem logical to me, but to be refined

_No response_

_No response_

danielmursa-dev:
Fixes #550 

- Implemented cache for objects
- Added env var for timeout


src/objects/conf/base.py
```diff
@@ -34,6 +34,16 @@
...
 TIME_ZONE = "Europe/Amsterdam"
 
+#
+# Caches
+#
+
+OBJECTTYPE_VERSION_CACHE_TIMEOUT = config(
+    "OBJECTTYPE_VERSION_CACHE_TIMEOUT",
+    default=5 * 60,  # 300 seconds
+    help_text="Timeout for cache when retrieving objecttype versions.",
```
stevenbal, line 44:
```suggestion
    help_text="Timeout in seconds for cache when retrieving objecttype versions.",
```

src/objects/core/utils.py
```diff
@@ -1,37 +1,48 @@
...
+
 
 def check_objecttype(object_type, version, data):
-    client = build_client(object_type.service)
-    objecttype_version_url = f"{object_type.url}/versions/{version}"
+    @cache(
+        f"objecttypen-{object_type.uuid}:versions-{version}",
+        timeout=settings.OBJECTTYPE_VERSION_CACHE_TIMEOUT,
+    )
+    def get_objecttype_version_response():
```
stevenbal, line 16: The implementation works, but I think it would be nice if we have a similar implementation as in Open Forms:

https://github.com/open-formulieren/open-forms/blob/7f5390b52572da99dbcf5c2430244e260ec098a4/src/openforms/contrib/objects_api/clients/objecttypes.py#L63

Basically an `ObjecttypesClient` class which implements `get_objecttype_version` (uncached) and another method that calls `get_objecttype_version` but is cached. If we keep the structure the same, we might be able to create a library at some point for all these common clients

src/objects/tests/v2/test_validation.py
```diff
@@ -19,7 +19,7 @@
 
 
 @requests_mock.Mocker()
-class ObjectTypeValidationTests(TokenAuthMixin, APITestCase):
+class ObjectTypeValidationTests(TokenAuthMixin, ClearCachesMixin, APITestCase):
```
stevenbal, line 22: Could you add a test to verify that creating two objects with the same type within the timeout causes only one request to be done and when creating an object after the cache key has expired the request is done again?

stevenbal (changes requested):
Looks good overall :+1:

src/objects/tests/v2/test_validation.py
```diff
@@ -31,6 +31,61 @@ def setUpTestData(cls):
...
+            self.assertEqual(Object.objects.count(), 5)
+
+        with self.subTest("clear_cache"):
+            m.reset_mock()
+            self.assertEqual(m.call_count, 0)
+            for n in range(5):
+                self._clear_caches()
+                self.client.post(url, data, **GEO_WRITE_KWARGS)
+            self.assertEqual(m.call_count, 5)
+            self.assertEqual(Object.objects.count(), 10)
```
stevenbal, line 70: I'd prefer a test here that doesn't explicitly clear the cache, but instead uses `freezegun.freeze_time`:

```python
with freeze_time(<some_timestamp>):
  # create two objects
  # assert only one GET request was done
  # reset mock

with freeze_time(<some_timestamp+timeout>):
  # create another object
  # assert one GET request was done

```

stevenbal (changes requested):

src/objects/tests/v2/test_validation.py
```diff
@@ -31,6 +35,74 @@ def setUpTestData(cls):
...
+
+        with self.subTest("cache_timeout"):
+            m.reset_mock()
+            old_datetime = datetime.datetime(2025, 5, 1, 12, 0)
+            with freeze_time(str(old_datetime.time())):
```
stevenbal, line 79: I think `freeze_time` accepts ISO 8601 formatted datetimes as well
```suggestion
            with freeze_time(old_datetime.isoformat()):
```

src/objects/tests/v2/test_validation.py
```diff
@@ -31,6 +35,74 @@ def setUpTestData(cls):
...
+                self.assertEqual(m.call_count, 0)
+                self.client.post(url, data, **GEO_WRITE_KWARGS)
+                self.client.post(url, data, **GEO_WRITE_KWARGS)
+                # only one request for two post
+                self.assertEqual(m.call_count, 1)
+
+            # cache_timeout is still ok
+            cache_timeout = settings.OBJECTTYPE_VERSION_CACHE_TIMEOUT
+            new_datetime = old_datetime + datetime.timedelta(0, (cache_timeout - 60))
```
stevenbal, line 88:
```suggestion
            new_datetime = old_datetime + datetime.timedelta(seconds=(cache_timeout - 60))
```

stevenbal (approved):
Minor test stuff, looks good otherwise

PR #577, opened 2025-04-29
fix: added missing type for ObjectRecord - data field in openapi.yaml

### LINKED ISSUE #576: ObjectRecord - data field in the OpenAPI spec is invalid - misses type property

3.0.3

The ObjectRecord - data field in the OpenAPI spec is invalid - misses type property. For this reason we cannot generate client code from this OpenAPI spec because it is invalid.

Very easy to fix in the spec itself. I will create a PR for this. Hopefully this is all there is to it. 

_No response_

_No response_

edgarvonk:
Added missing type for ObjectRecord - data field in openapi.yaml so that client code can successfully be generated from this OpenAPI specification

Fixes #576

**Changes**

- Added missing type for ObjectRecord - data field in openapi.yaml so that client code can successfully be generated from this OpenAPI specification

### DIFF
--- a/src/objects/api/v2/openapi.yaml
+++ b/src/objects/api/v2/openapi.yaml
@@ -1016,6 +1016,8 @@ components:
           minimum: 0
           description: Version of the OBJECTTYPE for data in the object record
         data:
+          type: object
+          additionalProperties: {}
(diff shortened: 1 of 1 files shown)


Floris272:
Will be fixed in #584.

PR #581, opened 2025-05-06
Issue/570 remove geo widget

### LINKED ISSUE #570: Geometry in the Objects API admin doesn't seem to work consistently

Version unknown (versionnr not shown anymore?)

Discussed during standup, the Objects API doesn't seem to load the geometry anymore when adding/editting an object. CSP errors or other changes?

_No response_

_No response_

Floris272:
Fixes #570 

**Changes**
- removes geodadminMixin & sets the widget for geometry fields to TextArea
- Fixes issue added in #571 
   - `check_objecttype_cached` from `core/utils` is called in by the admin & the api, the errors  return "non_field_errors" as the key, which DRF uses by default. 
   - The admin however only allows fields of the model or `"__all__"`.

### DIFF
--- a/src/objects/core/admin.py
+++ b/src/objects/core/admin.py
@@ -1,7 +1,8 @@
 import logging
 
 from django.contrib import admin
-from django.contrib.gis.admin.options import GeoModelAdminMixin
+from django.contrib.gis import forms
+from django.contrib.gis.db.models import GeometryField
 from django.http import JsonResponse
 from django.urls import path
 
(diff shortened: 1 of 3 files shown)


src/objects/core/admin.py
```diff
@@ -62,6 +63,8 @@ class ObjectRecordInline(GeoModelAdminMixin, admin.TabularInline):
         "correct",
     )
 
+    formfield_overrides = {GeometryField: {"widget": forms.Textarea}}
```
stevenbal, line 66: Do we know what format you have to use to enter coordinates? Might be good to have an example in the help text if we do

stevenbal (approved):

danielmursa-dev (approved):

PR #582, opened 2025-05-12
:bug: [#374] Fix: Token auth select is empty when creating a permission

### LINKED ISSUE #374: UI Issue when creating a permission
When creating a permission, the token auth foreign key keeps showing the empty `-------` option:

image

This seems to be a UI issue as the instance does have the FK set correctly.

OlhaZahoruiko:
Fixes #374


stevenbal (approved):

stevenbal:
@Floris272 I'll leave merging of this to you, you can include it in 3.0.4 if you want to

PR #585, opened 2025-05-14
:bug: changed default backend

TimdeBeer1:
Fixes maykinmedia/open-klant#414

**Changes**
Changed default filter backend and used commonground-api-common branch:
https://github.com/maykinmedia/commonground-api-common/pull/94

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -59,7 +59,7 @@ click-plugins==1.1.1
     # via celery
 click-repl==0.3.0
     # via celery
-commonground-api-common==2.6.2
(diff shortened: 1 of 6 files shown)


stevenbal (commented):
When a new version of `commonground-api-common` is released, you can remove the git dependency from `base.in` and run `bin/compile_dependencies.sh -P commonground-api-common==...` to upgrade to that new version.

Also might be better to change the commit message to mention that you upgraded the library to version 2.x.x

danielmursa-dev (approved):

PR #626, opened 2025-06-18
:bug: [#619] fix duplicate logs and only show structured logs

### LINKED ISSUE #619: The Celery container still outputs some unstructured logs

3.1.0

Having deployed the latest version to get structured logs, I'm still seeing some un-structured logs being emitted in the celery worker container, together with structured logs. I suspect log propagation is causing things to be logged twice.

Example:

```
2025-06-06 15:01:32,538 INFO celery.worker.consumer.connection  Connected to redis://objecten-maykin-test-redis:6379/1
{"event": "Connected to redis://objecten-maykin-test-redis:6379/1", "source": "app", "timestamp": "2025-06-06T13:01:32.539903Z", "logger": "celery.worker.consumer.connection", "level": "info"}
```

When logging in structured mode, I do *NOT* want to see unstructured logs, especially if the log events are identical.

1. Pull image tag 3.1.0 - this should be able to reproduce with docker compose as well
2. Inspect the container logs with `docker logs <container>`

* Logs are not duplicated
* Structured logs are emitted only

TimdeBeer1:
Fixes #619 

**Changes**

Fixed duplicate logs
only shows structured logs


src/objects/celery.py
```diff
@@ -82,6 +82,21 @@ def receiver_setup_logging(
                     "handlers": ["console"],
                     "level": "INFO",
                 },
+                "celery": {
```
stevenbal, line 85: Hmm this does make the configuration diverge quite a bit from Open Forms, which should work the same and doesn't add these extra loggers (https://github.com/open-formulieren/open-forms/blob/master/src/openforms/celery.py#L76).

Could you check if Open Forms also does not output duplicate logs and see what the configuration difference is that causes this issue in Objects?

stevenbal (changes requested):

PR #650, opened 2025-08-04
:sparkles: [open-api-framework#157] add sql data dump script

Floris272:
Fixes #https://github.com/maykinmedia/open-api-framework/issues/157

**Changes**
- Adds data dump script

### DIFF
--- a/bin/dump_data.sh
+++ b/bin/dump_data.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+
+# Dump component data
+# This script is not intended for a data migration to another Objects Api instance.
+# Run this script from the root of the repository
+#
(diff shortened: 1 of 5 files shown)


bin/dump_data.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+
+# Dump component data
+# This script is not intended for a data migration to another Objects Api instance.
+# Run this script from the root of the repository
+#
+# Note that postgres 17 requires postgres-client-17
```
stevenbal, line 7: I forgot, but is it possible to dump a postgres <17 database with postgres-client-17? Or is it not backwards compatible?

PR #652, opened 2025-08-13
✨ [maykinmedia/open-api-framework#159] add model img docs creation

TimdeBeer1:
Fixes maykinmedia/open-api-framework#159

**Changes**

Added functionality to create objects model img and automatically add it to the docs.

Added objecttypes manually


docs/uml_directive/uml.py
```diff
@@ -0,0 +1,34 @@
+import os
+
+from docutils import nodes
+from docutils.parsers.rst import Directive
+
+
+class UmlImagesDirective(Directive):
```
stevenbal, line 7: @TimdeBeer1 would it be possible to use the updated directive in `commonground-api-common` here as well? Or are there some customizations needed here?

docs/introduction/uml-diagram.rst
```diff
@@ -0,0 +1,24 @@
...
+
+.. note::
+
+    These are the underlying data models and this shows the relationships between the resources,
+    but not all attributes are the exact same as in the API.
+
+Objects API
+-----------
+.. uml_images::
+    :apps: core
```
stevenbal, line 14: Could you exclude the `Service` model? This is more of an internal thing that is not really interesting from an API perspective
afbeelding

stevenbal (changes requested):
Looks good, just a comment about excluding the `Service` model

PR #658, opened 2025-08-28
:sparkles: [#653] add created_on & modified_on to object & record models

### LINKED ISSUE #653: Show created/modified datetime of Object and Record
### Thema / Theme

Admin

Currently I only see a date via the admin, which makes debugging potential duplicated objects quite complicated.

Please add created_on/modified_on fields on both the Object and the Object Record and expose these via the admin. This should making dealing with large numbers of objects a little bit more manageable 

_No response_

_No response_

Floris272:
Fixes #653 

**Changes**
- adds `created_on` & `modified_on` to object and record.


src/objects/api/serializers.py
```diff
@@ -43,13 +43,17 @@ class Meta:
             "registrationAt",
             "correctionFor",
             "correctedBy",
+            "createdOn",
+            "modifiedOn",
```
stevenbal, line 47: In the issue it was only mentioned to expose these via the admin, so let's stick to that for now
```suggestion
```

src/objects/api/serializers.py
```diff
@@ -79,13 +83,17 @@ class Meta:
...
```
stevenbal, line 87: Same remark, let's only expose these in the admin for now

stevenbal (changes requested):

stevenbal (approved):

PR #662, opened 2025-09-09
:zap: [#661] Add GINIndex on ObjectRecord.data

### LINKED ISSUE #661: Slow requests on objects list when filtering on data_attrs

3.1.4 / latest

From Utrecht taiga (ref. 398)

The Commutr migration has been postponed due to Objects API having slow response times for GETs on `/objects`. In particular the following requests are slow:

  - Slow count query `/api/v2/objects?type=/api/v2/objecttypes/b5b47406-ee38-4fec-ba8b-6f2326e772a2&pageSize=1&page=1&data_attrs=inwonerprofielId__exact__<uuid>&ordering`
  - Slow count and slow base query `/api/v2/objects?type=/api/v2/objecttypes/ca2b8f4b-950b-4bbd-b0be-1f05cae5d875&pageSize=1000&page=1&data_attrs=zaaknummer__exact__<zaaknummer>&ordering=-record__data__contactmoment__datumContact`

When performing single requests, the response time was around 800ms, but when doing multiple in parallel this quickly increased to 5s+, possibly due to these requests triggering full table scans. The database had around 171k objectrecords.

Will try to reproduce this locally, I suspect that with enough objects (100k+) that have data with identifiers and having one that matches a specific value, I will be able to reproduce this locally

_No response_

stevenbal:
Fixes #661

Doesn't seem to affect the response times in CI too much, this only starts having effect when there is a decent number of objects in the database (100k+), because that's when full table scans start being expensive.

Locally I tested this with a million objectrecords, and there the count query goes from 256ms to 3ms due to the index being used

I tested the effect of the index on inserts, with a `gin_pending_list_limit` of 4MB (postgres default) I noticed about up to 6 inserts out of 10000 that were slow (100ms to 600ms). When testing without the index, this occurred less often (0 or 1 out of 10000 were slower than 100ms).

**Changes**

* Add GINIndex on ObjectRecord.data and change filters to use `__contains` to utilize this index


danielmursa-dev (approved):
👍

PR #675, opened 2025-09-24
Remove maxcdn from error templates

Floris272:
**Changes**

`500.html` & `503.html` used external bootstrap styling for the button, but since CSP is blocking it anyway it has been removed.


Floris272:
will be fixed in https://github.com/maykinmedia/objects-api/pull/665

PR #678, opened 2025-09-26
:zap: [#677] Denormalize ObjectRecord by adding _object_type

### LINKED ISSUE #677: GET on objects without data_attr filters bad performance

3.2.0

Taiga Utrecht 403

On a dataset with 3mil objects, 3.6mil objectrecords a query that filters only on objecttype returns about 20k objects. This results in extremely slow queries

_No response_

_No response_

stevenbal:
Fixes #677

Reduces the amount of joins by denormalizing `_object_type` on `ObjectRecord`. Indexes have also been added to improve performance.

A data migration was added to backfill this denormalized column, which took about 1 hour to complete locally for 3.8 million ObjectRecords

Main: https://github.com/maykinmedia/objects-api/actions/runs/18089517249/job/51466666860#step:8:646
After changes: https://github.com/maykinmedia/objects-api/actions/runs/18124724855/job/51577259367?pr=678#step:8:620

**Changes**

* Denormalize ObjectRecord by adding _object_type and add ind indexes to benefit from this denormalized column
* Use non superuser token and add perftest for objecttype filter

### DIFF
--- a/src/objects/core/migrations/0033_objectrecord__backfill_denormalized_fields.py
+++ b/src/objects/core/migrations/0033_objectrecord__backfill_denormalized_fields.py
@@ -6,6 +6,7 @@ import threading
 from concurrent.futures import ThreadPoolExecutor
 from time import sleep
 
+from django.conf import settings
 from django.db import connection, migrations, transaction
 
 from structlog import get_logger
@@ -15,7 +16,11 @@ logger = get_logger(__name__)
 
 
 BATCH_SIZE = int(os.getenv("OBJECTRECORD_MIGRATION_0033_BATCH_SIZE", 2_000))
-NUM_WORKERS = int(os.getenv("OBJECTRECORD_MIGRATION_0033_NUM_WORKERS", 4))
+NUM_WORKERS = int(os.getenv("OBJECTRECORD_MIGRATION_0033_NUM_WORKERS", 2))
+
+# Pooling does not work with multiple workers
+if "pool" in settings.DATABASES["default"].get("OPTIONS", {}):
+    NUM_WORKERS = 1
 
 
 def backfill_object_type_batch_concurrent(cursor):
@@ -49,7 +54,7 @@ def worker(apps, progress):
     Worker that keeps grabbing batches until none are left.
     """
     # Stagger the workers to avoid synchronized bursts of commit I/O
-    delay = random.uniform(0.5, 1.5)
+    delay = random.uniform(0.1, 0.3)
     sleep(delay)
 
     while True:
@@ -78,12 +83,14 @@ def forward(apps, schema_editor):
     progress = tqdm(
         total=total_records, desc="Backfilling ObjectRecords", smoothing=0.1
     )
+    if NUM_WORKERS == 1:
+        worker(apps, progress)
+    else:
+        with ThreadPoolExecutor(max_workers=NUM_WORKERS) as executor:
+            futures = [executor.submit(worker, apps, progress) for _ in range(NUM_WORKERS)]
 
-    with ThreadPoolExecutor(max_workers=NUM_WORKERS) as executor:
-        futures = [executor.submit(worker, apps, progress) for _ in range(NUM_WORKERS)]
-
-        for f in futures:
-            f.result()
+            for f in futures:
+                f.result()
 
 
 class Migration(migrations.Migration):
--- a/src/objects/core/tests/test_migrations.py
+++ b/src/objects/core/tests/test_migrations.py
@@ -3,57 +3,7 @@ import threading
 import time
 from unittest.mock import patch
 
-from django.core.management import call_command
-from django.db import connection
-from django.db.migrations.executor import MigrationExecutor
-from django.db.migrations.state import StateApps
-from django.test import TransactionTestCase
-
-
-# TODO move this to maykin common?
-class BaseMigrationTest(TransactionTestCase):
-    app: str
-    migrate_from: str  # The migration before the one we want to test
-    migrate_to: str  # The migration we want to test
-
-    setting_overrides: dict = {}
-
-    old_app_state: StateApps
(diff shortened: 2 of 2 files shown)


src/objects/core/migrations/0033_objectrecord__backfill_denormalized_fields.py
```diff
@@ -0,0 +1,44 @@
...
+    total = 0
+    last_id = 0
+
+    while True:
+        batch = ObjectRecord.objects.filter(id__gt=last_id).order_by("id")[:BATCH_SIZE]
+        ids = list(batch.values_list("id", flat=True))
+        if not ids:
+            break
+
+        ObjectRecord.objects.filter(id__in=ids).update(
```
sdegroot, line 25: A single batch of documents in our database takes about 18 seconds. With a total of 3 million, this operation would take about 90 minutes. That is a bit much for a database migration. Since this migration is nothing more than copying data from one table to another, would it be an idea to simply run a `insert into select from`?

sdegroot (changes requested):

src/objects/core/migrations/0033_objectrecord__backfill_denormalized_fields.py
```diff
@@ -0,0 +1,48 @@
...
+logger = get_logger(__name__)
+
+
+BATCH_SIZE = int(os.getenv("OBJECTRECORD_BATCH_SIZE", 200_000))
+
+
+def forward(apps, schema_editor):
+    with connection.cursor() as cursor:
+        # Get min and max id in core_objectrecord
+        cursor.execute("SELECT MIN(id), MAX(id) FROM core_objectrecord;")
```
sdegroot, line 18: This migration is run while other pods may actively insert new objects. This could lead to a situation where not all rows are actually updated

src/objects/core/migrations/0034_alter_objectrecord__object_type_and_more.py
```diff
@@ -0,0 +1,31 @@
...
+
+    dependencies = [
+        ('core', '0033_objectrecord__backfill_denormalized_fields'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='objectrecord',
+            name='_object_type',
+            field=models.ForeignKey(blank=True, help_text='OBJECTTYPE in Objecttypes API', null=True, on_delete=django.db.models.deletion.PROTECT, to='core.objecttype'),
```
sdegroot, line 17: You initially set the field to `nullable` but after the migration there should not be any null values anymore, right?

src/objects/core/migrations/0033_objectrecord__backfill_denormalized_fields.py
stevenbal: changed it to filter by _object_type = null instead

src/objects/core/migrations/0033_objectrecord__backfill_denormalized_fields.py
```diff
@@ -0,0 +1,50 @@
...
+          AND r.object_id = o.id;
+    """,
+        [batch_size],
+    )
+    return cursor.rowcount
+
+
+def forward(apps, schema_editor):
+    with connection.cursor() as cursor:
+        while True:
```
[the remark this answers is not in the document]
stevenbal: Hmm yeah that's true, if this makes it into a release I'll make sure there's a warning mentioning this

src/objects/core/migrations/0033_objectrecord__backfill_denormalized_fields.py
```diff
@@ -0,0 +1,50 @@
...
+logger = get_logger(__name__)
+
+
+BATCH_SIZE = int(os.getenv("OBJECTRECORD_BATCH_SIZE", 200_000))
+
+
+def backfill_object_type_batch(apps, cursor, batch_size):
+    cursor.execute(
+        """
+        WITH batch AS (
```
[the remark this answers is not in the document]
stevenbal: With the index it seems to be a little bit faster, so I've changed it to add the index before the data migration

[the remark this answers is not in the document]
sdegroot: @stevenbal I do not see the index in the source code, is that the visualisation here or is it indeed still applied afterwards? 

In production it takes about 5 minutes per batch. There is no index on objectrecord._object_type_id and therefore it has to do full table scans on the whole table (3.6 million records). Note that the json data in each record is about 1kb per record. Since postgres is a row database, without indexes, it will load the whole row in memory, thus having to scan about 4 GB of data for every iteration.

The migration is now running almost an hour and we have set the maximum to 90 minutes (we hoped that would be enough) :(

src/objects/core/models.py
```diff
@@ -154,11 +154,35 @@ class ObjectRecord(models.Model):
         auto_now=True, help_text=_("Last modification date")
     )
 
+    # Denormalized field to avoid unnecessary joins on `Object`
+    _object_type = models.ForeignKey(
+        ObjectType,
+        on_delete=models.PROTECT,
```
danielmursa-dev: @stevenbal Isn't this a bit forced? I mean, this way, if you want to delete the object type, you have to check two instances, one for Object (because it has the same protection) and one for `ObjectRecord` before deleting it, and in any case, if you set `SET_NULL` here, you put `self._object_type = self.object.object_type` in `save()`, so it gets updated automatically every time.

stevenbal: We can't use `SET_NULL` here, because `null` has been set to `False` after the datamigration. This is needed to make sure that we know for sure that each ObjectRecord has a `_object_type`, because that is the column we're filtering on

danielmursa-dev: Okok thanks!

danielmursa-dev (approved):

PR #693, opened 2025-10-29
:arrow_up: [maykinmedia/open-api-framework#191] upgrade nodejs to v24

TimdeBeer1:
Fixes maykinmedia/open-api-framework#191

**Changes**

Upgraded nodejs to v24

### DIFF
--- a/src/objects/js/components/admin/permissions/index.js
+++ b/src/objects/js/components/admin/permissions/index.js
@@ -1,15 +1,15 @@
-import React from "react";
-import ReactDOM from "react-dom";
+import React from "react"; 
+import { createRoot } from "react-dom/client";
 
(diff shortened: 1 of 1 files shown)


stevenbal (changes requested):
It seems that the permission form isn't working correctly after the upgrade (`/admin/token/permission/<pk>/change/`), could you look into this?

afbeelding

afbeelding

TimdeBeer1:
> It seems that the permission form isn't working correctly after the upgrade (`/admin/token/permission/<pk>/change/`), could you look into this?
> afbeelding afbeelding

In React 18+ ReactDOM.render is deprecated. Now it can be done like this: https://react.dev/reference/react-dom/client/createRoot

PR #713, opened 2025-12-24
Feature/564 open objecten

### LINKED ISSUE #564: OpenObjecten - combine the ObjectsType API into the Objects API
### Thema / Theme

Objecten API

From #550 

The idea is to add the ObjectTypes API into the Objects API, making use of the same database. This to avoid requests between the two (improves performance) and  simplify the architecture and deployments. Discussed with Utrecht and Den Haag

Initial estimation:

> For OpenObjecten: the migration path from Objects API+ObjectTypes API will be quite complicated, as we don't see a workable path to convert the Objects API to OpenObjecten (this would require calling the OT API during datamigrations which feels hacky). So the best alternative would be to create a new OpenObjecten component and add a management command to migrate the data from an existing ObjectsAPI+ObjectTypes API to the new OpenObjecten component: 2-3 months of work.

DH mentions that a migration would be workable if the API paths remain intact, then the objecttypes-api domainname would simply point to the objects-api domainname.

#### TODO:
- [ ] Move the fields on objecttypes.ObjectType model to objects.ObjectType (nullable/blank for now?)
  - [ ] also merge the admins for this model (perhaps move the new fields to a separate fieldset?)
- [ ] move the v2 API endpoints to Objects
- [ ] Add a management command that:
  - [ ] takes as input base URL for objecttypes api, token for objecttypes API
  - [ ] retrieve all objecttypes (with pagination) and create local objecttype instances from them
    - [ ] if a local ObjectType with that UUID exists already: update it, otherwise create a new one
  - [ ] Document how to use this management command 
- [ ] in ``check_objecttype_cached`` add an extra code path to use the local ObjectType schema if it exists, if not it should keep the current behavior (retrieving and caching the objecttype info from objecttypes API). This way it's still backwards compatible
- [ ] make a minor release that includes the above changes
- [ ] use django_upgrade_check to ensure objecttypes are migrated before upgrading to 4.0?
  - [ ] not sure if this is possible with django_upgrade_check, but it would be nice if we could check if all local ObjectTypes have been correctly migrated
- [ ] in version 4.0, remove the old code path and make fields required (where necessary, should be the same as it was before in the objecttypes.ObjectType model)
- [ ] Remove any references to Objecttypes API as a separate component in the documentation
- [ ] Remove the external Objecttypes API from the docker-compose setup

#### To be discussed:
- [ ] I'm assuming both the API endpoints will be under `/api/v2/`? E.g. `/api/v2/objects/` and `/api/v2/objecttypes/`

Improved performance
Simplification of architecture
Simplification of deployments

Floris272:
Fixes #564 

**Changes**
- fully integrates objecttypes api into objects api
- removes code used for fetching external objecttypes
- adds upgrade check to make sure all objecttypes are imported

TODO
- [ ] docs. install & readme

### DIFF
--- a/src/objects/core/management/commands/import_objecttypes.py
+++ b/src/objects/core/management/commands/import_objecttypes.py
@@ -1,4 +1,3 @@
-
 from django.core.management.base import BaseCommand, CommandError
 from django.db import transaction
 from django.utils.translation import gettext as _
--- a/src/objects/setup_configuration/steps/token_auth.py
+++ b/src/objects/setup_configuration/steps/token_auth.py
@@ -1,4 +1,3 @@
-
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db import IntegrityError
 


src/objects/tests/v2/test_validation.py
```diff
@@ -1,217 +1,43 @@
...
-                "startAt": "2020-01-01",
-            },
-        }
-
-        response = self.client.post(url, data, **GEO_WRITE_KWARGS)
-
-        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
-        self.assertEqual(Object.objects.count(), 0)
-
-    def test_create_object_with_invalid_length(self, m):
```
stevenbal: Shouldn't this test be kept? Or is it no longer applicable?

src/objects/api/v2/openapi.yaml
```diff
@@ -816,6 +816,462 @@ paths:
               schema:
                 $ref: '#/components/schemas/PaginatedObjectList'
           description: OK
+  /objecttypes:
```
stevenbal: Because of this new endpoint, the API_VERSION should get a minor version bump

docker/setup_configuration/data.yaml
```diff
@@ -57,14 +36,15 @@ tokenauth:
...
       administration: Administration 1
-      permissions:
-        - object_type: b0e8553f-8b1a-4d55-ab90-6d02f1bcf2c2
-          mode: read_only
-          use_fields: true
-          fields:
-            '1':
-              - record__data__leeftijd
-              - record__data__kiemjaar
+# TODO
```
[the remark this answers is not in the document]
stevenbal: What's the reason for removing it? Is it because you can't create ObjectTypes now without specifying the JSON schema?

stevenbal (changes requested):
Looks good overall :+1:

stevenbal (changes requested):
Some small things remaining still, I'd recommend to split off the duplicate UUID stuff into a separate PR, because this PR has been open for quite a while already

PR #727, opened 2026-02-05
:bug: [#718] Fix 500 error on duplicate UUID when creating objects

### LINKED ISSUE #718: POST to `/objecten` with existing UUID raises 500 error

3.5.0 / latest

When attempting to do a POST to `/objecten` with a UUID that is already used by another object, a 500 error is raised

```
IntegrityError at /api/v2/objects
duplicate key value violates unique constraint "core_object_uuid_key"
DETAIL:  Key (uuid)=(289f65bc-e1c0-43f4-b2c4-e71181315f45) already exists.
```

1. Do a POST to `/objecten` to create an Object
2. Do a second POST to `/objecten` and reuse the UUID from step 1
3. Observe the internal server error

In any case, Objects API should raise a 400 error a duplicate UUID is used. 

I'm wondering though if it's intentional (or desirable) to let users specify UUIDs, this might have been an oversight?

OlhaZahoruiko:
Fixes #718

**Changes**

[Describe the changes here]

### DIFF
--- a/src/objects/api/serializers.py
+++ b/src/objects/api/serializers.py
@@ -3,6 +3,7 @@ from django.utils.translation import gettext_lazy as _
 
 import structlog
 from rest_framework import serializers
+from rest_framework.validators import UniqueValidator
 from rest_framework_gis.serializers import GeometryField
 
 from objects.core.models import Object, ObjectRecord, ObjectType, Reference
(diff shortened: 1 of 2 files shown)


src/objects/api/serializers.py
```diff
@@ -125,12 +125,35 @@ class Meta:
         }
         validators = [JsonSchemaValidator(), GeometryValidator()]
 
+    def validate(self, attrs):
```
stevenbal: I think we could make use of `rest_framework.validators.UniqueValidator` here (https://www.django-rest-framework.org/api-guide/validators/#uniquevalidator), you can add it to the `validators` on `uuid`.

Make sure to double check if creating a new version for an Object still works (I assume this is part of the unittests already though)

stevenbal (approved):

PR #737, opened 2026-02-26
:memo: [#564] refactor docs for objecttype migrations

### LINKED ISSUE #564: OpenObjecten - combine the ObjectsType API into the Objects API

Floris272:
Fixes #564

**Changes**
- updates docs

### DIFF
--- a/docs/manual/migration.rst
+++ b/docs/manual/migration.rst
@@ -3,7 +3,7 @@
 ObjectTypes API migration
 =========================
 
-In version 4.0.0 the ObjectTypes API will be merged into the Objects API so that only one application is needed (Open Objecten).
+In version 4.0.0 the ObjectTypes API will be merged into the Objects API so that only one application is needed (Open Object).
 This means that from version 4.0.0 onwards only objecttypes that exist in the database are supported, and no external objecttypes can be used.
 
 Version flow
--- a/Dockerfile
+++ b/Dockerfile
@@ -93,7 +93,7 @@ RUN python src/manage.py collectstatic --noinput
 LABEL org.label-schema.vcs-ref=$COMMIT_HASH \
       org.label-schema.vcs-url="https://github.com/maykinmedia/objects-api" \
       org.label-schema.version=$RELEASE \
-      org.label-schema.name="objects API"
+      org.label-schema.name="Open Object"
 
 EXPOSE 8000
 CMD ["/start.sh"]
--- a/README.NL.rst
+++ b/README.NL.rst
@@ -1,5 +1,5 @@
 =============
-Open Objecten
+Open Object
 =============
 
 :Version: 3.6.0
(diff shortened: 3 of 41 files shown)


docs/admin/object.rst
```diff
@@ -62,7 +61,7 @@ corrected in the "Correction" field of the next record.
 .. image:: _assets/img/object_create_record.png
     :alt: Add an object record
 
-In the Objects API you always see one record, which contains data of a certain time (by default
+In the API you always see one record, which contains data of a certain time (by default
```
stevenbal:
```suggestion
In the API you always see one record for a specific object, which contains data of a certain time (by default
```

docs/manual/migration.rst
stevenbal:
```suggestion
In version 4.0.0 the ObjectTypes API will be merged into the Objects API so that only one application is needed (Open Objecten). This means that from version 4.0.0 onwards only objecttypes that exist in the database are supported, and no external objecttypes can be used.
```

PR #769, opened 2026-06-04
:sparkles: [#518] Add searchable objectTypes selector and sort alphabetically

### LINKED ISSUE #518: I want to be able to search for objectTypes in permission view dropdown
### Thema / Theme

Objecten API

For the sake of completion and maybe tactical to fix them all together;
Currently 3 issues with the permission page:
1. Auth not set; https://github.com/maykinmedia/objects-api/issues/374
2. Slow loading of permissions page; DH Taiga 596 (Should a github ticket be created?)
3. I want to be able to search for objectTypes; This ticket

In DH we already have quite some object Types. When I use the admin UI to add a permission i get a lengthily dropdown where I have to find the desired objectType in a seemingly random order.
Quick fix: At least sort them aphabetically
Actual request: make them searchable or something so that I can quickly find the desired permission.
Currently I am having cases where I have to add 10 permissions to an auth (for every environment) and have to aimlessly scroll through a huge list finding the type.

Bonus: Multiselect so I dont have to go through this screen multiple times, though this is less of an issue if all 3 previously mentioned issues are resolved.

Saving a lot of time and 'inconveniences'.

_No response_

OlhaZahoruiko:
Fixes #518

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

- Architectural design record

  - [x] If a design decision was made that changes functionality, create an architectural design record for it [here](https://github.com/maykinmedia/open-object/wiki/Architectural-Decision-Records-(ADR))

### DIFF
--- a/src/objects/js/components/admin/permissions/permission-form.js
+++ b/src/objects/js/components/admin/permissions/permission-form.js
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from "react";
 
 import { CheckboxInput, TextInput, SelectInput } from "../../forms/inputs";
 import { versionAuthFields } from "./auth-fields";
+import {SearchableSelectInput} from "../../forms/searchable-select-input";
 
 
 const PermissionForm = ({objectFields, tokenChoices, objecttypeChoices, modeChoices, formData}) => {
@@ -72,7 +73,7 @@ const PermissionForm = ({objectFields, tokenChoices, objecttypeChoices, modeChoi
         </div>
 
         <div className="form-row">
-            <SelectInput
+            <SearchableSelectInput
                 name="object_type"
                 id="id_object_type"
                 label="Object type:"
--- a/src/objects/js/components/forms/searchable-select-input.js
+++ b/src/objects/js/components/forms/searchable-select-input.js
@@ -1,0 +1,106 @@
+import React, { useState } from "react";
+import Select from "react-select";
+
+import { ErrorList } from "./error-list";
+
+const styles = {
+    container: (base) => ({
+        ...base,
+        maxWidth: "30rem",
+    }),
+    control: (base) => ({
+        ...base,
+        backgroundColor: "var(--body-bg)",
+        color: "var(--body-fg)",
+        borderColor: "var(--hairline-color)",
+    }),
+    menu: (base) => ({
+        ...base,
+        backgroundColor: "var(--body-bg)",
+        color: "var(--body-fg)",
+    }),
+    option: (base, state) => ({
+        ...base,
+        backgroundColor: state.isFocused
+            ? "var(--selected-row)"
+            : "var(--body-bg)",
+        color: "var(--body-fg)",
+    }),
+    singleValue: (base) => ({
+        ...base,
+        color: "var(--body-fg)",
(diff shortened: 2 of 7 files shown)


src/objects/token/tests/test_admin.py
```diff
@@ -41,3 +41,16 @@ def test_with_object_types_api_v2(self):
             choices[1][1],
             str(object_type),
         )
+
+    @tag("#518")
```
stevenbal:
```suggestion
    @tag("gh-518")
```

src/objects/js/components/forms/searchable-select-input.js
```diff
@@ -0,0 +1,102 @@
+import React, { useState } from "react";
```
stevenbal, line 1: Although open-api-framework is currently only a python dependency, I *think* we might be able to move this file to that library and then import from open-api-framework in `permission-form.js`, that would save us some copy pasting when we want to add this widget elsewhere.

Could you check if that's possible?

src/objects/js/components/admin/permissions/permission-form.js
```diff
@@ -72,7 +73,7 @@ const PermissionForm = ({objectFields, tokenChoices, objecttypeChoices, modeChoi
         </div>
 
         <div className="form-row">
-            <SelectInput
+            <SearchableSelectInput
```
stevenbal, line 76: The widget is a bit wide when using fullscreen, could you make it a bit less wide?

Image

stevenbal (changes requested):

src/objects/js/components/admin/permissions/permission-form.js
OlhaZahoruiko: Screenshot from 2026-06-05 13-11-47
Does this size look OK to you?

src/objects/js/components/forms/searchable-select-input.js
OlhaZahoruiko: I checked this, but it doesn't seem possible at the moment. I tried importing a JavaScript module from OAF

```js
import { TestWidget } from 'open_api_framework/static/open_api_framework/js/test-widget';
```

and webpack failed with

```text
Module not found: Error: Can't resolve 'open_api_framework/static/open_api_framework/js/test-widget'
```

It looks like OAF is currently only available as a Python dependency and doesn't expose JavaScript modules that webpack can consume :confused:

src/objects/js/components/admin/permissions/permission-form.js
stevenbal: Yes, looks good :+1:

src/objects/js/components/forms/searchable-select-input.js
stevenbal: Hmm alright, thanks for checking. I don't feel like setting up an NPM package for something as small as this either, so let's just keep it as it is now. If we need it in other projects, we can copy it to there as well

stevenbal (approved):

PR #774, opened 2026-06-08
:sparkles: [#731] Jsonschema format validation

### LINKED ISSUE #731: Schema should validate custom format attributes & update data_attrs filter
### Thema / Theme

Objecten API

See parent issue

#### jsonschema
https://python-jsonschema.readthedocs.io/en/latest/validate/#validating-formats
- add `jsonschema[format-nongpl]` dependency
- add ` format_checker=Draft202012Validator.FORMAT_CHECKER` to jsonschema.validate
- add custom email check
- add tests for all formats
- add strict boolean to ObjectTypeVersion model which should defines if the format_checker should be used 
  - should default to config setting

#### data_attrs API filter
- add datetime, time & duration, (maybe delta) to `utils.string_to_value`
- other fields are strings and should work with existing `exact` & `icontains` filters

_No response_

_No response_

danielmursa-dev:
Fixes #731 

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene

  - [ ] Commit messages refer to the relevant Github issue
  - [ ] Commit messages explain the "why" of change, not the how

- Architectural design record

  - [ ] If a design decision was made that changes functionality, create an architectural design record for it [here](https://github.com/maykinmedia/open-object/wiki/Architectural-Decision-Records-(ADR))


stevenbal (changes requested):
@danielmursa-dev could you rebase on master to fix the merge conflicts?

danielmursa-dev:
> @danielmursa-dev could you rebase on master to fix the merge conflicts?

@stevenbal Done

src/objects/conf/base.py
```diff
@@ -267,3 +267,11 @@
         help_text=("Defines the primary domain where the application is hosted."),
     ),
 )
+
+JSONSCHEMA_USE_FORMAT_CHECKER = config(
+    "JSONSCHEMA_USE_FORMAT_CHECKER",
+    default=True,
+    documentation=DocumentationParams(
+        help_text=("Enable JSON Schema format validation."),
```
stevenbal:
```suggestion
        help_text=("Whether or not JSON Schema format validation for objects is enabled by default. This can be overridden per object type."),
```

src/objects/core/migrations/0040_objecttypeversion_strict_format_checker.py
```diff
@@ -0,0 +1,29 @@
...
+    operations = [
+        migrations.AddField(
+            model_name="objecttypeversion",
+            name="strict_format_checker",
+            field=models.BooleanField(
+                help_text="When enabled, JSON Schema validation enforces format constraints Defaults to the JSONSCHEMA_USE_FORMAT_CHECKER setting.",
+                null=True,
+            ),
+        ),
+        migrations.RunPython(set_default_false, migrations.RunPython.noop),
```
stevenbal: I guess to rely less on runpython, you could first have addfield with default=False and then just change the default to `None`?

Having this and the JSONSCHEMA_USE_FORMAT_CHECKER by default True does mean that someone has to manually set each objecttype to True to enable it everywhere though. But inverting it means that manual work is required to disable it, so there's not really a way around that I guess. Let's make sure it's properly mentioned in the changelog though

src/objects/core/tests/test_jsonschema.py
```diff
@@ -0,0 +1,313 @@
...
+
+from jsonschema import FormatError
+
+from ..utils import (
+    is_valid_color,
+    is_valid_email,
+    validate_jsonschema,
+)
+
+SCHEMA_ALL_FORMATS = {
```
stevenbal, line 12: Given that https://github.com/open-zaak/open-zaak/issues/2355 is also going to use jsonschema, could we move this and other code in this PR to open-api-framework?

src/objects/core/tests/factories.py
```diff
@@ -26,7 +26,7 @@ class ObjectTypeVersionFactory(factory.django.DjangoModelFactory[ObjectTypeVersi
         {
             "type": "object",
             "title": "Tree",
-            "$schema": "http://json-schema.org/draft-07/schema#",
+            "$schema": "https://json-schema.org/draft/2020-12/schema#",
```
stevenbal, line 29: Does this schema version change affect object creation in any other way?

src/objects/core/tests/test_objecttypeversion_generate.py
```diff
@@ -44,3 +49,112 @@ def test_version_bounds_check(self):
...
+            version = ObjectTypeVersionFactory.create()
+            self.assertFalse(version.strict_format_checker)
+
+            version.strict_format_checker = True
+            version.save()
+            self.assertTrue(version.strict_format_checker)
+
+        with override_settings(JSONSCHEMA_USE_FORMAT_CHECKER=False):
+            version = ObjectTypeVersionFactory.create(strict_format_checker=True)
+            self.assertTrue(version.strict_format_checker)
```
stevenbal, line 67: These two seem to test the same behavior?

I think a test for `strict_format_checker=None` is implied by the create without args, but I'd like to see that explicitly specified as well

src/objects/core/models.py
```diff
@@ -203,6 +211,9 @@ def clean(self):
         check_json_schema(self.json_schema)
 
     def save(self, *args, **kwargs):
+        if self.strict_format_checker is None:
+            self.strict_format_checker = settings.JSONSCHEMA_USE_FORMAT_CHECKER
```
stevenbal: This works a little bit different from what I expected, I assumed we would simply store `None` and then use the value of `settings.JSONSCHEMA_USE_FORMAT_CHECKER` at runtime. Then again I'm not really against this, I'm only wondering though if it actually needs to be a nullable booleanfield at this point, I guess we could also remove `null=True` and then have default be some callable that returns `settings.JSONSCHEMA_USE_FORMAT_CHECKER`? This means that we have fewer possible values to worry about (if for some reason a `null` value makes it into the database).

I think not having the null option and using the default also simplifies the admin form a bit, because it will just preselect the value of settings.JSONSCHEMA_USE_FORMAT_CHECKER, and not have three options

src/objects/core/models.py
```diff
@@ -183,6 +184,13 @@ class ObjectTypeVersion(models.Model):
     json_schema = models.JSONField(
         _("JSON schema"), help_text=_("JSON schema for Object validation"), default=dict
     )
+    strict_format_checker = models.BooleanField(
```
stevenbal, line 197: Let's also add this to the changelist `list_display` for the admin, and maybe add a filter as well to easily find objecttypes that have this enabled/disabled

stevenbal (changes requested):
@danielmursa-dev could you also add at least one test with failing format validation via the API?

src/objects/conf/base.py
stevenbal: I think when we release this, the changelog should also make it clear that existing objecttypes have this disabled

src/objects/core/tests/factories.py
danielmursa-dev: I don't think so, because all the tests pass

src/objects/core/tests/test_jsonschema.py
danielmursa-dev: Yes exactly this was a point, because the same code is in `open-vtb` and `open-product` as well.
Can we do it in this PR or in another one?

src/objects/conf/base.py
danielmursa-dev: Done, in CHANGELOG :+1:

src/objects/core/tests/test_objecttypeversion_generate.py
danielmursa-dev: Fixed :+1:

src/objects/core/models.py
danielmursa-dev: Good!

stevenbal: Could you add it to `list_filter` as well?

src/objects/core/tests/test_jsonschema.py
stevenbal: Let's do it in a separate one

src/objects/core/models.py
stevenbal: Hmm I still think that I would prefer changing to a regular BooleanField (without allowing None), I think it simplifies things a bit and leaves less room for error if there happen to be rows with `None` in the database somehow. Also it means we don't have to customize the widget, which currently shows "Unknown":

Image

You can do something like:

```python
strict_format_checker = models.BooleanField(default=lambda: settings.JSONSCHEMA_USE_FORMAT_CHECKER)
```

Do you disagree?

CHANGELOG.rst
```diff
@@ -2,6 +2,20 @@
 Change history
 ==============
 
+4.2.0 (XXXX-XX-XX)
+------------------
+
+.. note::
+
+    A new ``JSONSCHEMA_USE_FORMAT_CHECKER`` setting has been introduced (default: ``True``) for newly created objecttypes.
```
stevenbal:
```suggestion
    A new ``JSONSCHEMA_USE_FORMAT_CHECKER`` setting has been introduced (default: ``False`` for existing objecttypes when upgrading to 4.2.0, default: ``True`` for newly created objecttypes after that).
```

stevenbal (changes requested):

src/objects/core/models.py
danielmursa-dev: Done!

stevenbal (approved):

